PR #142, opened 2018-03-23
Fetch non-included tax adjustments in Order getTotalTaxablePrice

efatsi:
I have 2 separate custom tax adjusters in a plugin, and the second one is factoring in the first's when calculating a taxable price. Looks like the `getTotalTaxablePrice` method tries to mitigate this by subtracting out the `$taxAdjustmets`, but unfortunately it filters for those that are `included`.

I'm _pretty sure_ this behavior is unexpected and that it should be fetching non-included tax adjustments.

### DIFF
--- a/src/elements/Order.php
+++ b/src/elements/Order.php
@@ -401,7 +401,7 @@ class Order extends Element
         $itemTotal = $this->getItemSubtotal();
 
         $allNonIncludedAdjustmentsTotal = $this->getAdjustmentsTotal();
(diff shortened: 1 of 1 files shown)


lukeholder:
Hi @efatsi, thanks for reporting. The method should really return the total minus the tax adjustments and the included tax adjustments; the real total of items without any taxes. I have made that commit, and will close this PR. 

Let me know if my change makes sense to you: 484ae39ae9becf30ce

efatsi:
Nice, thanks for digging in there. Looks great.

Issue #136, opened 2018-03-21
Can't add variants


davidhellmann:
![image](https://user-images.githubusercontent.com/764292/37741432-a9cf95fa-2d61-11e8-9cb2-ed512a1d4530.png)

Console:
`Uncaught SyntaxError: Invalid or unexpected token`

Details: 
```
[...]

PR #143, opened 2018-03-26
Fix Bug that the variants-container is hidden. #136

davidhellmann:
#136 related


lukeholder:
Thanks for for the fix!

PR #329, opened 2018-04-23
Mimic structure of sitemenu from other Craft elements

davist11:
The sitemenu button on other Craft element entry screens has the class of `sitemenubtn` and then each site item link within the menu has the appropriate `data-site-id`. This mimics that functionality so that we can more easily target this menu consistently with JS between Craft and Commerce.


lukeholder:
Thanks!

PR #336, opened 2018-05-02
Add _returnCart to RemoveAllLineItems

jordikroon:
This PR adds a `_returnCart` method call to the actionRemoveAllLineItems.

The _returnCart method saves the cart, thus by only calling setLineItems the cart doesn't update in the end. This results in a broken action.


lukeholder:
Thanks!

krizajb:
When will this be live?

lukeholder:
This is now available in 2.0.0-beta.5

PR #348, opened 2018-05-18
Return redirect result in PaymentsController#actionPay

efatsi:
Think I've encountered a bug attempting to implement a custom payment gateway. On successful authorization/payment/whatever, the last thing that `actionPay` does is define the redirect target. However it then goes on to return `null`, seemingly invalidating the defined redirect.

This patch seems to do the trick for me locally in having the app redirect to the specified `$order->returnUrl` when it's available.


lukeholder:
Thanks, this is the correct way to complete the requests.

PR #372, opened 2018-06-12
Add Product column to Variant index table

### LINKED ISSUE #333: Variant Modal picker
### Description
Can we display the product title as well as the variant title in the variant model please.

If you allow to show variant titles, then you don't really know which product it belongs to. ANd if you have multiple products with variants that use the same naming format, then there is no way to tell the difference

### Additional info

- Craft version: Craft Pro 3.0.4
- PHP version: 7.1.16
- Database driver & version: MySQL 5.5.5-10.2.14-MariaDB
- Plugins & versions:

AugustMiller:
Fixes #333 by adding a product column to `defineTableAttributes`, `defineDefaultTableAttributes`, and a case in the switch statement within `getTableAttributeHtml`.

~This does not make any change to how elements are rendered in the CP—this seems to be governed primarily by one function in the View component.~

~I may open an issue in the cms repo suggesting this be broken out a bit, such that we could more easily override just the contents of an element chiclet, but not have to reconstruct it entirely (or parse the HTML in a second hook, make modifications, and pack back into a string).~

### DIFF
--- a/src/elements/Variant.php
+++ b/src/elements/Variant.php
@@ -726,6 +726,7 @@ class Variant extends Purchasable
     {
         return [
             'title' => Craft::t('commerce', 'Title'),
+            'product' => Craft::t('commerce', 'Product'),
             'sku' => Craft::t('commerce', 'SKU'),
             'price' => Craft::t('commerce', 'Price'),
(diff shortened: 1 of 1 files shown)


lukeholder:
Good addition. Thanks.

PR #374, opened 2018-06-17
Remove blank element from `getStatesAsList` output

AugustMiller:
A blank array element was inserted at the beginning of each Country group. I wouldn't consider this the expected output—seems like the template's responsibility to insert empty `option` elements, not to filter them out of what ought to be a clean representation of the data.

This PR eliminates the empty entry.

### DIFF
--- a/src/services/States.php
+++ b/src/services/States.php
@@ -110,11 +110,6 @@ class States extends Component
 
         foreach ($states as $state) {
             $cid2state += [$state->countryId => []];
-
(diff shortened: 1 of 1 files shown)


lukeholder:
Do you see any errors with this change in the example templates when setting a custom stateValue? Or in the order edit page in an address edit modal?

lukeholder:
Tested and doesn't look like this causes any issues. Thanks!

PR #429, opened 2018-07-27
Add ability to customize fields captured in snapshots

AugustMiller:
In a project's `commerce.php` config file, merchants/developers can provide two new keys:
1. `variantSnapshotFields`
2. `productSnapshotFields`

…both can be set to an empty array to disable any content field fetching, or to select just a subset of available fields. Each element in the array must be a field's handle.

This applies to both Variants and Products. A default `null` value preserves current functionality.

### DIFF
--- a/src/elements/Variant.php
+++ b/src/elements/Variant.php
@@ -341,10 +341,12 @@ class Variant extends Purchasable
         $data['product'] = $this->getProduct() ? $this->getProduct()->getSnapshot() : [];
 
         // Variant Custom Field values
-        $data['productFields'] = $this->getProduct() ? $this->getProduct()->getSerializedFieldValues() : [];
+        $productFields = Plugin::getInstance()->getSettings()->productSnapshotFields;
+        $data['productFields'] = $this->getProduct() ? $this->getProduct()->getSerializedFieldValues($productFields) : [];
 
         // Variant Custom Field values
-        $data['fields'] = $this->getSerializedFieldValues();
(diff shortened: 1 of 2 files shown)


lukeholder:
Thanks for this, will look to merge it soon after review.

lukeholder:
Thinking about this - the issue I have is that fields names are product type dependent.

So thinking a breaking change of no longer including the snapshot data is the right move here.

So we wouldn't go with a config, but go with an event that allows a plugin/module to modify the field array passed to `getSerializedFieldValues` so it also can have knowledge about the purchasable in question.

AugustMiller:
Great point—I hadn't considered this.

I may have some time to contribute, but it may require a deeper understanding of the underlying serialization strategy and field layouts?

…

On second thought, it's still just an array of field handles—so long as the event receives the `product` and a `fields` array as props, it seems like a pretty clean addition.

Because this is top-of mind, I'd like to do some hacking. I'll push anything that feels good, but as a different PR—nothing relevant here, with that approach. Feel free to close this!

AugustMiller:
☝️ Issued a separate PR for an event-based solution!

AugustMiller:
Another argument for making this backwards-compatible (and asking developers to opt-in to the performance improvement): it's not just Commerce 2 users we're potentially leaving in the lurch—it's anyone who relies on snapshots in a Commerce 1 install… otherwise, we're requiring that, after the update, the developer must determine whether new line items will have the data available, and either create a plugin (to manually include fields in the manifest), or modify their templates.

Just a thought!

PR #437, opened 2018-08-03
Event-based Variant and Product Snapshot Customization

AugustMiller:
This is accomplished by a series of two events, each for Variants and Products:
- Before the data is captured, there's an opportunity to modify the fields that will be gathered.
- After the data is captured, there's an opportunity to modify the data that comes back from the serialization.

This is a non-breaking change, as the default `null` value is preserved, if no plugins modify the field list.

### DIFF
--- a/src/elements/Variant.php
+++ b/src/elements/Variant.php
@@ -11,6 +11,10 @@ use Craft;
 use craft\base\Element;
 use craft\commerce\base\Purchasable;
 use craft\commerce\elements\db\VariantQuery;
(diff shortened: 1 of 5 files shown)


lukeholder:
Thanks for this addition - Implementation looks good.

I added change-log notes here: 6d69ddd5dcf589494e89125781

lukeholder:
Decided to force performance with a breaking change so: a5fb08c1763ad4d2dc6d59b1ffea976a2b1cba69

AugustMiller:
Thank you! 🙌

PR #445, opened 2018-08-15
Allow partial updates to address records

AugustMiller:
This allows you to do things like set only the `makePrimaryBillingAddress`, provided the `address[id]` param, which previously failed validation, because not all existing values might not be sent in a request.

In addition, you can update addresses piecemeal, by sending only the params you wish to update (instead of building a new Model and saving over the old one, each time).

This does not seem to affect existing implementations in the admin, or front-end—passing the full address is still acceptable.

Note: Changes in #444 are close by, so there might be some collisions. 😢


lukeholder:
Hi @AugustMiller keen to merge this if you can clean up conflicts with latest develop. Thanks.

AugustMiller:
Not great at this, but I think it's been brought up to date! @lukeholder

lukeholder:
Thanks!

PR #446, opened 2018-08-16
Point to correct class for `OrderStatusException`

AugustMiller:
For some reason, this wasn't producing an error at compile time—but when the error was actually used (Line 554), it blew up.

All this does is corrects the namespace/path of the class, and places it in alphabetic order.


lukeholder:
Thanks!

PR #448, opened 2018-08-16
Fix SQL error when saving Order Statuses

### LINKED ISSUE #447: Error when updating Order Status
### Description

When updating an existing order status with at least one Email selected, a foreign key constraint error is produced

### Steps to reproduce

1. Create an _Email_;
2. Create an _Order Status_;
3. Edit the _Order Status_ from Step 2;
4. Attach an Email to the _Order Status_;
5. Save the _Order Status_.

This should produce the SQL error, suggesting that the join table `commerce_orderstatus_emails` doesn't contain an expected row?

### Additional info

- Craft version: 3.0.20
- PHP version: 7.1.16
- Database driver & version: PostgreSQL 10.3
- Plugins & versions: Commerce 2.0.0-beta.7

AugustMiller:
The `$record` variable in `craft\commerce\services\OrderStatuses::saveOrderStatus` was being re-assigned inside a `foreach` loop, and causing some unexpected behavior, described in the referenced issue.

Fixes #447!

### DIFF
--- a/src/services/OrderStatuses.php
+++ b/src/services/OrderStatuses.php
@@ -216,10 +216,10 @@ class OrderStatuses extends Component
 
             //Delete old links
             if ($model->id) {
-                $records = OrderStatusEmailRecord::find()->where(['orderStatusId' => $model->id])->all();
+                $orderStatusEmailRecords = OrderStatusEmailRecord::find()->where(['orderStatusId' => $model->id])->all();
 
-                foreach ($records as $record) {
-                    $record->delete();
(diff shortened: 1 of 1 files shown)


dhahn:
I believe this can be simplified to a single query (and line) as:

```
 OrderStatusEmailRecord::deleteAll(['orderStatusId' => $model->id]);
```

AugustMiller:
That's a nice, terse way of handling it—not super familiar with ActiveRecord!

This completely side-steps the original issue, avoids unnecessary assignment, and (presumably) reduces the total number of database calls, in many cases. 👌

PR #469, opened 2018-08-29
Add ability to retrieve the cheapest variant from a product.

### LINKED ISSUE #468: Add ability to retrieve the cheapest variant from a product
It is a common requirement to get the cheapest variant from a product e.g. for displaying something like "price from $x" in the top level product listings.

This can be achieved using Twig, however based on a conversation I had about this in Slack, it seems as though there are quite a few people getting around it with Twig macros or similar which leads me to believe the addition would be heavily used and thus justified to add directly to the [`Product`](https://github.com/craftcms/commerce/blob/develop/src/elements/Product.php) element.

MartyWallace:
Resolves #468

### DIFF
--- a/src/elements/Product.php
+++ b/src/elements/Product.php
@@ -164,6 +164,11 @@ class Product extends Element
      */
     private $_defaultVariant;
 
+	/**
+	 * @var Variant This product's cheapest variant
+	 */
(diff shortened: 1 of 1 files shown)


lukeholder:
Thanks

PR #479, opened 2018-09-06
Split custom emails by commas into an array

efatsi:
Docs say you can `select "Send to custom recipients", which allows you to add any email addresses seperated by a comma`. Currently though, there's no parsing of this string so the "to" attribute of the email is set to the single string of comma separated emails, which bombs out as an invalid email address.

This PR splits the incoming `to` field by whitespace/comma and passes the resulting array in.


lukeholder:
thanks

PR #490, opened 2018-09-17
Update deprecation note for getTotalTax()

akshayagarwal:
The documentation wrongly implied using the 'taxIncluded' adjustment type for even the tax not included adjustment.


lukeholder:
Thanks!

Issue #422, opened 2018-07-18
On add to cart a LineItem doesn't have any adjusters because lineItemId is null


nfourtythree:
### Description
On add to a lineItem doesn't have a tax adjuster associated with it (The tax is on the whole order though).

I tracked it down to https://github.com/craftcms/commerce/blob/develop/src/adjusters/Tax.php#L163

When it hits this part on first add `$item->id` returns `null`. If you subsequently update the cart the tax adjuster starts working.

Here is a little gif example, it starts having just added the product to the cart and shows 0 for `item.getAdjustmentsTotalByType( 'tax', true )`: http://x.n43.me/2B1n0z2E021T/Screen%20Recording%202018-07-18%20at%2005.54%20pm.gif

(gif isn't frozen in the middle just my local vagrant box decided to be super slow on that submit)

### Steps to reproduce

1. Create tax that applies to line items and is included
2. Add to basket
3. Output `item.getAdjustmentsTotalByType( 'tax', true )` in `cart.lineItems` loop and it displays 0
4. Update the cart (forcing a recalculation) and it will start showing tax

### Additional info

- Craft version: 3.0.15
- PHP version: 7.1
- Plugins & versions: Commerce 2.0.0 beta 6

RitterKnightCreative:
I think this is related but I just ran into this wondering why my line items weren't showing up.

Here's a common use case...

After adding a new item to the cart where the customer has already put a coupon code in, Commerce will set the adjuster to be a cart adjustment and not an item adjustment.  

This is a problem because there's not really a good way to show "You saved X" on the individual line item if it's a cart adjustment.   The total discount coupon is right but it will look WRONG to a customer.

Updating the cart again immediately fixes the problem.

Here's a debug cart to illustrate... 

I just added a 10 lb bag (of flour) to the cart:

![screen shot 2018-10-22 at 10 30 27 am](https://user-images.githubusercontent.com/3707869/47302028-06ea1a80-d5e6-11e8-8b32-f49ea2dc15ee.png)

Updating the cart fixes the problem:

![screen shot 2018-10-22 at 10 30 56 am](https://user-images.githubusercontent.com/3707869/47302179-706a2900-d5e6-11e8-9822-81bc34379022.png)

Sounds like there just needs to be some internal update/cleanup happening.

joshangell:
I’ve just run into this as well ... it seems like if the line items haven’t yet got an ID they need saving first before the adjustments run. Or the order needs recalculating at some point.

[...]

PR #498, opened 2018-09-20
Fix for saving a line item with a negative total

dhahn:
When a line item's adjustments makes a the total for that line item negative saving
that line item errors out. This uses the same methodology that is used
for the sale price in that model.

Here is the SQL error:
```
SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'total' at row 1
```

You can replicate this by adding a custom adjuster that adds a discount greater than the salePrice of the item.

### DIFF
--- a/src/models/LineItem.php
+++ b/src/models/LineItem.php
@@ -291,7 +291,7 @@ class LineItem extends Model
      */
     public function getTotal(): float
     {
-        return $this->getSubtotal() + $this->getAdjustmentsTotal();
+        return max($this->getSubtotal() + $this->getAdjustmentsTotal(), 0);
     }
 
     /**


dhahn:
I can confirm this was previously not an issue in Craft Commerce 2.0.0-beta8.1, and it came about when we upgraded to beta10.

dhahn:
Actually, it looks like this has been an issue prior to 8.1. We were just unintentionally avoiding it locally by skipping custom line item adjustments for line items that didn't have an id.

lukeholder:
In effect this makes adjustments impossible to understand at the order level. i.e get total of discount adjustments, since it is ignoring some of those discount's values.

It would be better for the adjusters themselves to know if they are making the line item price less than zero. 

We need to think more about this, like making the discount adjuster aware of putting line items into negative.

dhahn:
@lukeholder I agree that there is likely a better solution. I think the root cause of this is that each individual adjustment is not aware of how much is left to be discounted. One major thing I could see that would need to be solved for something like this to happen is #422. To me the scariest piece would be having to figure out each custom adjusters fits into a process like this. I would almost rather there be another column on the line item that is allowed to go negative, but not used in calculations for payments. Either that or a column on the adjustments that is "total applied" which represents exactly how much of the discount adjustment was used to zero out the line item balance.

in the meantime, is there something you would suggest doing when writing custom discount adjusters to avoid this issue? It's not just custom adjusters either. If you were to set the "Per Item Discount" to something greater than the salePrice of any matched item you would receive the same error `SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'total' at row 1` as well.

dhahn:
Any update on this?

You can very easily replicate this if you add a discount code that discounts $100 from an order and then only add a $10 item to your cart.

lukeholder:
For now we are allowing negative line items as per e1d9b9026115f60199770c9c3355d1bf74a8b6f5 , it will be up to the payment controller to stop processing or completing negative orders. Happy to discuss the issue of negative order totals and how we can avoid them here #651

PR #499, opened 2018-09-20
Fix for associating adjustments to line items

dhahn:
Because new line items don't have an id yet, the `getAdjustments` method
on the line items model associates any base order adjustment with that new
line item. This adds a condition to avoid that weirdness.

This is very related to #498 and very much caused by #422

### DIFF
--- a/src/models/LineItem.php
+++ b/src/models/LineItem.php
@@ -485,7 +485,7 @@ class LineItem extends Model
         $adjustments = $this->getOrder()->getAdjustments();
 
         foreach ($adjustments as $adjustment) {
-            if ($adjustment->lineItemId == $this->id) {
+            if ($adjustment->lineItemId && $adjustment->lineItemId == $this->id) {
                 $lineItemAdjustments[] = $adjustment;
             }
         }


dhahn:
I can confirm this was previously not an issue in Craft Commerce 2.0.0-beta8.1, and it came about when we upgraded to beta10.

lukeholder:
This is still an issue, but not sure how your fix correct it?

If the line item has a null ID because it is not saved yet, the adjustment’s line item ID will be null, so I am not sure if the adjustment will be related to the right line item with your additional condition?

dhahn:
Yeah I think my description is a bit misleading. This doesn't associate the adjustment to the correct line item it just makes sure that all adjustments not associated to a line item are not associated to the new line item. The error still exists that an adjustment won't be associated to a line item until the next cycle when the line item has an id.

If you consider the case where you have an adjustment that does not have a line item id (base order adjustment) those are all being associated to a newly added line item.

lukeholder:
This has now been fixed on the `develop` branch.

PR #512, opened 2018-10-01
Progress meter on checkout page breaks if website in subfolder

strategyagency:
Using `craft.app.request.url|slice(1)` will cause a mismatch if the site lives in a subfolder, using `craft.app.request.pathInfo` fixes this.

If your website URL is domain.com/subfolder/, craft will interpret the current checkout step as `subfolder/shop/checkout/address` instead of `shop/checkout/address`


lukeholder:
Thanks

Issue #330, opened 2018-04-25
Multi Add to cart forms don't add items if one quantity is 0


nfourtythree:
### Description
This is a lead on from #321 opening a new issue for easier tracking of the issue (as requested)

My Notes from the other thread:

With the update, you now can't submit a multi-add form that has rows with 0 `qty`. If you take the example code in the docs and tweak it 

```twig
<form method="POST">
  {{ csrfInput() }}
    <input type="hidden" name="action" value="commerce/cart/update-cart">

    {% for variant in product.variants %}
        <input type="hidden" name="purchasables[{{loop.index}}][id]" value="{{ variant.purchasableId }}">
        <input type="text" name="purchasables[{{loop.index}}][qty]" value="1">
        <input type="hidden" name="purchasables[{{loop.index}}][note]" value="1">
    {% endfor %}

    <input type="submit" value="Add all variants to cart">
</form>
```

Then change the first `qty` box to be 0. You get a cart error with the validation error being you can't submit something with a `qty` less than 1.

Which as a data validation rule is fine, but in practice would mean that building a project you would either need to disabled the 0 qty fields on submit, or strip them from the post etc.

It would seem like in a multi-add form it is likely that qty data submitted would be a range from 0 upwards.

I know my pull request for this originally (#322) was far too simple but it was a way of not letting the data get to the point of validation.

Hope this helps, let me know if you need any more info. I came across this as my add to cart stopped working after updating

### Steps to reproduce

1. Copy and paste code above within a product with variants
1. Run the page change the first qty box to 0
1. **Expected outcome** All products added with quantities should add
1. **Actual outcome** Nothing adds

### Additional info

- Craft version: Craft Pro 3.0.4
- PHP version: 7.1
- Database driver & version: MySQL
- Plugins & versions: Commerce 2.0.0-beta3

lukeholder:
Submitting a zero qty is not valid, thus the error is raised. 

Your PR suggests we should silently ignore invalidly submitted data?

nfourtythree:
@lukeholder yes but in the case of a multi-add form there are going to be many quantity boxes. That is the nature of a form of that type.

For example in a form like this
![](http://x.n43.me/0y1i3D1E021s/Image%202018-04-26%20at%207.50.25%20am.png)

So without adding some JS, or hooking into a pre-add PHP event there isn't a way natively to stop those 0 qty rows adding.

bossanova808:
[...]

PR #522, opened 2018-10-09
Do not add items without an explicity set qty to the cart

bossanova808:
This is more compatible with Commerce V1 (and MultiAdd) behaviour, and is more explicit - only items with both an id AND a set qty should be added to a cart.

It allows for e.g. checkboxes to more easily be used in add to cart forms (for e.g. for upsells), so that what is submitted might look like:

```
action | commerce/cart/update-cart
purchasables[0][id] | 1385
purchasables[0][qty] | 1
purchasables[2][id] | 4700
purchasables[2][qty] | 1
purchasables[3][id] | 1302
purchasables[4][id] | 4711
```
...where the intention is clearly that only `1385` and `4700` are added to the cart, and not the other two items.  This is pretty much a standard pattern in multi-item add to cart forms.

### DIFF
--- a/src/controllers/CartController.php
+++ b/src/controllers/CartController.php
@@ -173,7 +173,7 @@ class CartController extends BaseFrontEndController
         if ($purchasableId = $request->getParam('purchasableId')) {
             $note = $request->getParam('note', '');
             $options = $request->getParam('options') ?: [];
-            $qty = (int)$request->getParam('qty', 1);
+            $qty = (int)$request->getParam('qty', 0);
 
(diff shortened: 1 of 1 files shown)


lukeholder:
I thought this was the current behavior?

When passing 0 as the qty on mass line item updating, it removes the line item.

When adding to the cart, purchasable IDs with qty of zero are ignored based on discussion here: https://github.com/craftcms/commerce/issues/330#issuecomment-384533139

If we defaulted to zero this would be a breaking change. Couldn't you pass a qty value of zero or not include the purchasable ID? - this way the form is more explicit.

PR #525, opened 2018-10-13
Fix error in ShippingRules.php

engram-design:
I'm getting the following error (attached screenshot).

`Return value of craft\commerce\services\ShippingRules::getAllShippingRulesByShippingMethodId() must be of the type array, null returned`

This would seem a pretty easy fix, as noted by the functions return type.

screen shot 2018-10-14 at 12 12 27 am

### DIFF
--- a/src/services/ShippingRules.php
+++ b/src/services/ShippingRules.php
@@ -80,7 +80,7 @@ class ShippingRules extends Component
         }
 
         if ($this->_fetchedAllShippingRules) {
-            return null;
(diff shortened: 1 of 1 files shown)


lukeholder:
Not needed, remnant of shipping method service copy pasta 2ec2cf7ed6cc95bc96f38ade62f16bf3f0c284f6

engram-design:
Even better, thanks @lukeholder

PR #528, opened 2018-10-16
Fix default checked value

watarutmnh:
Shipping and billing primary checkboxes default values are seem to be expected opposite. Are't they?


lukeholder:
Thanks!

PR #556, opened 2018-11-07
Update available-variables.md

nilsenpaul:
This input needs to have `shippingMethodHandle` as name, or else it will not save the corresponding handle.


lukeholder:
Thanks!

PR #570, opened 2018-11-15
Fixed typos causing the multiple purchases code to throw errors

MattWilcox:
each line had a mis-matched closing ]


angrybrad:
Thanks!

PR #571, opened 2018-11-16
Add `EVENT_AFTER_REMOVE_LINE_ITEM` event for Order

engram-design:
Just wanted to add this event for compatibility with [Instant Analytics](https://github.com/nystudio107/craft-instantanalytics) but also just handy to have.


lukeholder:
Thanks.

PR #572, opened 2018-11-16
force hasStock to be boolean

samuelbirch:
When using Craft.BaseElementSelectInput and passing in the criteria, hasStock is not treated as a boolean and therefore does not work.

I've forced a check on the variable when used in the query to fix this issue.


lukeholder:
Thanks.

PR #579, opened 2018-11-21
Sort order in orders widget

DiederikVanHoorebeke:
The orders were sorted by date ASC, which should be DESC, in the Recent Orders widget.


lukeholder:
Thanks!

PR #589, opened 2018-11-27
Ability to limit the minimum total sale threshold of an order

martywallace:
Currently when discounts are applied to an order, the total value of the order can go into a negative value e.g.

```
Line Item Value  $10.00
Shipping         $15.00
Discount         $30.00

Total            $-5.00
```

I haven't looked into the actual ramifications of this (e.g. what happens when you submit this to particular gateways) but regardless of that, the display on the frontend using conventional means and the order value display and reporting within the dashboard essentially show as owing customers money.

In my case I have a particular requirement where we actually always want to at least charge for the shipping, which complicates the solution a little further.

This probably isn't the ideal way to solve the problem but hopefully you can at least see what I'm trying to accomplish and we can iron out something more suitable. Essentially this gives the ability to choose between three methods of calculating the total sale value of an order:

1. Default - the current behaviour wherein the above scenario will result in a negative sale value.
2. Zero - orders must have a value of at least $0.00 (can't be below this).
3. Shipping - a third option for my requirement where orders must at least cost their shipping.

I guess even if this approach was suitable we should at the very least record the selected strategy with orders in-case the option is changed during the runtime of a production site.

Please let me know how to proceed with implementing this functionality or whether you are able to take it on yourself.


lukeholder:
Discussing this in addition to negative line items. Will have a solution and standard before the GA.

lukeholder:
The issue for discussion related to this proposal is located here: #651

martywallace:
Sorry @lukeholder missed this for some reason, but looks like you sorted it out. Do you need me to do anything from here?

lukeholder:
This PR has been implemented in c5e955911afeef3c5f5abac542f86ffadc6deb90

Due out in the next release.

PR #593, opened 2018-12-02
Rename `getShippingCategoriesByProductId` method and single call

AugustMiller:
Now `getShippingCategoriesByProductTypeId`, to reflect that it's fetching _Product Type_ -> Shipping Category relationships, not _Product_ -> Shipping Category relationships (for which there isn't a join table, and a service method isn't needed/available).

This appears to be a safe change (only reference was in `ProductType::getShippingCategories`), but might require a deprecation notice, in case someone is using this service method directly in their templates?

### DIFF
--- a/src/models/ProductType.php
+++ b/src/models/ProductType.php
@@ -202,7 +202,7 @@ class ProductType extends Model
     public function getShippingCategories(): array
     {
         if (!$this->_shippingCategories) {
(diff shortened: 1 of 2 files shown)


lukeholder:
The method name is wrong, this is a good fix. No need for depreciation warning as this will be included in the GA release and did not exist in Commerce 1.

Issue #583, opened 2018-11-26
Item notes causing table layout issues


lukeyouell:
### Description

The item note column compresses the item title column (pictured) - this could cause friction when managing orders. I'd suggest placing the item note in it's own full-width row.

![image](https://user-images.githubusercontent.com/24975058/49011389-24bf1800-f16e-11e8-8567-02644a26a77a.png)

lukeholder:
Fixed in Commerce 3 Alpha

PR #604, opened 2018-12-14
Table Notes + Messages Output

AugustMiller:
This moves line item notes and order status messages into tooltips, so the tables remain readable when long messages are included.

I believe this addresses the concern in #583!


lukeholder:
In 2.1 we will introduce a redo of the order edit page, and this is a good tempory fix until then.

Thanks again!

PR #605, opened 2018-12-15
Get Plans by Information Entry ID

AugustMiller:
This seemed like a useful method—we'll often have a plan's informational Entry tucked away on a page, and need to be able to do a reverse-lookup to render a subscription form.

In a template, this makes it possible to:

```twig
{% set plan = craft.commerce.plans.getPlansByInformationEntryId(entry.id) | first %}
```

I figured returning an array was safer, granted you _can_ attach the same Entry to multiple plans…

### DIFF
--- a/src/services/Plans.php
+++ b/src/services/Plans.php
@@ -97,17 +97,7 @@ class Plans extends Component
         $results = $this->_createPlansQuery()
             ->all();
 
-        $plans = [];
-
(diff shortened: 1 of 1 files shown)


AugustMiller:
Important to note that this includes a non-trivial DRY-up of the multi-row Plan model population!

PR #612, opened 2018-12-24
Solve code error in setting up variable

mufaddalsmart1:


mufaddalsmart1:
You're most welcome. :)

PR #624, opened 2019-01-10
This fixes #621: use correct table name for Customer DiscountUses.

### LINKED ISSUE #621: customer_discountuse Table does not exist
In src/services/Discounts.php, line 535, `commerce_customer_discountuse` is wrong. This tables doesn't exist.

TheFox:


lukeholder:
Thanks for the Pull Request, we will be consistent for now and use the string version of the table name. If Craft CMS core switches to the tableName() we also will do so.

Fixed in 8ba2331

PR #639, opened 2019-01-21
Bugfix: Removing all items from cart (Commerce Lite)

luke-nehemedia:
Fixes an issue where removing the last item from the cart would cause an error:

> Undefined offset: 0 in _/craft/vendor/craftcms/commerce/src/elements/Order.php_

This would only happen in the Lite-Version, since it is limited to one item in the cart.

Checking at first, whether `$lineItems` has at least one item, before running `array_slice()`, fixes the issue.


lukeholder:
Thanks. This was fixed in 94aa300ead3d1f4b9ad12f474a5c8ede424346e9

PR #640, opened 2019-01-22
Change Dutch translation for 'Sale' to 'Aanbieding'

nilsenpaul:
Right now, Commerce uses two different Dutch words for the English word 'Sale'. In my opinion, both words don't cut it. The proper word would be 'aanbieding', as is also currently used for the English word 'promotions'.


angrybrad:
Thank you!

PR #642, opened 2019-01-22
Add the App helper for use on line 538

nilsenpaul:
An error gets thrown after changing the URI format for a product type.


lukeholder:
Thanks, this was fixed in ce684161da6a80727b580562e22f148f8281bf38

PR #663, opened 2019-01-28
Review and update French Canadian

jesuismaxime:
- Use the right apostrophe character used in French
- Use the right translation for the word "Order(s)" and its fellow adjective and past participle
- Use the right translation for the word "Tax(es)" 
- Corrected various syntax errors


lukeholder:
Thanks!

PR #664, opened 2019-01-28
 Make order statuses translatable in sidebar and CP columns

jesuismaxime:


lukeholder:
The 'commerce' translation category is for Commerce plugin supplied strings.

Since the order status is supplied by the user, we can't add translation to the status name in the UI. Take a look at using static translation strings: https://docs.craftcms.com/v2/static-translations.html

Thanks

PR #669, opened 2019-01-29
Add more info to errors thrown for emails

engram-design:
Currently dealing with an odd issue (on production, but not on local), which I'm trying to troubleshoot. 

Error reported in logs: `2019-01-30 09:34:12 [-][16244][-][error][craft\commerce\services\Emails::sendEmail] Email “Order Received Notification” could not be sent for order “f78a094”. Error: Creating default object from empty value`

Now, there's a lot of content in my emails, so that error is quite annoying in practice. This PR adds to the full file path to the template, along with the line number the error occurred on.

**Note:** my particular issue points to `barrelstrength/sprout-email/src/services/SentEmails.php:72` - but without this additional logging info, I wouldn't have been able to figure that out.

### DIFF
--- a/src/services/Emails.php
+++ b/src/services/Emails.php
@@ -331,10 +331,12 @@ class Emails extends Component
 
                 $newEmail->setTo($emails);
             } catch (\Exception $e) {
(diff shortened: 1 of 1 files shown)


lukeholder:
Thanks!

PR #684, opened 2019-02-05
Add a method to get the matching rule description

jesuismaxime:
Similar to `getPriceForOrder()`, but to make the description easily available (without a json_encode filter on the rules).

Useful to display some specifics related to a ShippingMethod to the end user.

### DIFF
--- a/src/base/ShippingMethod.php
+++ b/src/base/ShippingMethod.php
@@ -175,7 +175,7 @@ abstract class ShippingMethod extends Model implements ShippingMethodInterface
 
     /**
      * @param Order $order
-     * @return float
+     * @return string
      */
     public function getDescriptionForOrder(Order $order)
     {
--- a/src/base/ShippingMethod.php
+++ b/src/base/ShippingMethod.php
@@ -174,6 +174,21 @@ abstract class ShippingMethod extends Model implements ShippingMethodInterface
     }
 
     /**
+     * @param Order $order
+     * @return float
+     */
+    public function getDescriptionForOrder(Order $order)
+    {
+        $shippingRule = $this->getMatchingShippingRule($order);
+
+        if (!$shippingRule) {
+            return 0;
+        }
+
+        return $shippingRule->description;
+    }
+    
+    /**
      * @deprecated in 2.0
      */
     public function getAmount()
--- a/src/base/ShippingMethod.php
+++ b/src/base/ShippingMethod.php
@@ -182,7 +182,7 @@ abstract class ShippingMethod extends Model implements ShippingMethodInterface
         $shippingRule = $this->getMatchingShippingRule($order);
 
         if (!$shippingRule) {
-            return 0;
+            return;
         }
 
         return $shippingRule->description;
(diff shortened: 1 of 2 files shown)


ccchapman:
For what it's worth, I believe retrieving the shipping rule (whether the matching one only or all) for carts/orders is already possibly in Twig. (Without the `json_encode` filter, simply an example to show the object)

An example listing all of the shipping rules and their descriptions for a cart:

```
{% set cart = craft.commerce.cart %}
{% set shippingMethods = cart.availableShippingMethods %}

{% for shippingMethod in shippingMethods %}
  {% set shippingRules = shippingMethod.shippingRules %}

  {% for shippingRule in shippingRules %}
    {{ shippingRule.name }}
    {{ shippingRule.description }}
  {% endfor %}
{% endfor %}
```

src/base/ShippingMethod.php
```diff
@@ -173,6 +173,21 @@ public function getPriceForOrder(Order $order)
         return $amount;
     }
 
+    /**
+     * @param Order $order
+     * @return float
```
ccchapman: If added, the return type should be updated here and in the interface.

jesuismaxime:
From the official documentation :
```
{% for handle, method in cart.availableShippingMethods %}
    <label>
        <input type="radio" name="shippingMethodHandle" value="{{ handle }}"
               {% if handle == cart.shippingMethodHandle %}checked{% endif %} />
        <strong>{{ method.name }}</strong> {{ method.priceForOrder(cart)|currency(cart.currency) }}
    </label>
{% endfor %}
```

If `priceForOrder()` make the shipping amount available with no effort, there's no reason to hide the description under `shippingRules`. Description is useful to make a better feed-back for the front-end user (like the difference between Regular or Express shipping)

ccchapman:
Understand your desired functionality here.

Just want to be clear that it is already possible, and this PR simply creates a shortcut. I updated my first comment to provide another code example where you can get only the matching rule.

I will defer to the maintainers to determine if it creates overhead or if it's beneficial.

lukeholder:
The shipping method interface and base class will also need updating before we consider a merge. Also, please add the methods to the `shipping methods` docs page. Docs are now in the repo.

Thanks

bymayo:
I couldn't find any of this in the documentation (Regarding finding the shipping rules for a method), or it's pretty well hidden! Is it worth dropping this in the docs?

nfourtythree:
Hey @jesuismaxime 

As @ccchapman mentioned you can already return the matching rule for by using the `getMatchingShippingRule` method. This returns the rule which in turn has the `description`.

As you mentioned there is `getPriceForOrder` this is a bit more specific as it has to do some extra work in totaling the price etc.

I am not sure there would be much benefit in adding a specific method to just return the description. As this method would call `getMatchingShippingRule` anyway and then only return the description. Although it would be seen as more convenient it adds maintenance overhead and other things. Would also have to take into account that potentially someone might be returning a description that is blank and therefore you "truthy" tests would need to be more strict rather than testing against if the rule matched and then whether or not there is a description.

Issue #689, opened 2019-02-06
Return completed order object in OrdersController actionCompleteOrder()


ccchapman:
Curious if we could return the (already present) order object after an order has been successfully marked completed.

https://github.com/craftcms/commerce/blob/7eb5eff836b39169d2a79897d595a67478972a23/src/controllers/OrdersController.php#L320

Thoughts?

ccchapman:
On a related note, I am interested if the snapshot could be included for line items of completed orders.

ccchapman:
Realized that this method is not intended for public usage; will open a PR for a similar response when paying for the order.

PR #690, opened 2019-02-06
Add order object to actionCompleteOrder response

ccchapman:
Thinking this is a beneficial add-on to the complete order action response. Open to alternatives if there are concerns adding this.

Closes https://github.com/craftcms/commerce/issues/689

### DIFF
--- a/src/controllers/OrdersController.php
+++ b/src/controllers/OrdersController.php
@@ -315,7 +315,11 @@ class OrdersController extends BaseCpController
 
         if ($order && !$order->isCompleted && $order->markAsComplete()) {
             $date = new \DateTime($order->dateOrdered);
-            return $this->asJson(['success' => true, 'dateOrdered' => $date]);
+            return $this->asJson([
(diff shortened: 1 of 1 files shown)


ccchapman:
I noticed that the order should be passed through `BaseFrontEndController::cartArray()` before being made public in this case.

src/controllers/OrdersController.php
```diff
@@ -315,7 +315,11 @@ public function actionCompleteOrder(): Response
 
         if ($order && !$order->isCompleted && $order->markAsComplete()) {
             $date = new \DateTime($order->dateOrdered);
-            return $this->asJson(['success' => true, 'dateOrdered' => $date]);
+            return $this->asJson([
+                'success' => true,
+                'dateOrdered' => $date,
+                'order' => $order,
```
ccchapman: I noticed that the order should be passed through `BaseFrontEndController::cartArray()` before being made public in this case.

PR #693, opened 2019-02-07
Fix transaction status processing

boboldehampsink:
First check if its successful or processing before checking if its a redirect. This fixes some Mollie issues I have with payment methods where the status can be "open".


boboldehampsink:
This used to be the case but got messed up at a certain point: https://github.com/craftcms/commerce/commit/7bf0b26e4ea9872cca02846bcfc93873d384796b

andris-sevcenko:
Glad we're not point fingers, @boboldehampsink  :)

angrybrad:
☝ It was Andris. ☝

Issue #695, opened 2019-02-07
Rounding strategies and decimal place pricing


N0ps32:
### Description
Commerce incorrectly adds 1-2c to our order total even though it shouldn't.

### Steps to reproduce

1. Create products with the following prices:
Luftpolsterfolie (4.917 EUR)
Klebeband (1.667 EUR)
BOX (3.75 EUR)
Reisekoffer (8.25 EUR)
Business-Trolley (5.75 EUR)
Golftasche (12.417 EUR)

2. Add a 20%, not included in price VAT tax rate.

3. Add products to cart:
 3x BOX
 1x Reisekoffer
 1x Business-Trolley
 1x Golftasche
 1x Klebeband
 1x Luftpolsterfolie

4. Your cart.totalPrice will now be € 53.11 even though it should be € 53.10.
    Because ((3.75 * 3) + 8.25 + 12.417 + 5.75 + (1.667 + 4.917)) * 1.2 = **53.10**12

Screenshots:
https://i.imgur.com/kW5OSKG.png
https://i.imgur.com/DNPB0zX.png

This is a very big issue for us. We are essentially overcharging clients 2c that we would need to refund leading to much more complications.

### Additional info

- Craft version: Craft Pro 3.1.8
- PHP version: 7.2.13
- Database driver & version: MySQL 5.5.5
- Plugins & versions: Craft Commerce | 2.0.4, Stripe for Craft Commerce 1.0.10

ccchapman:
Your adjustments and line item totals seem to be correct in the screenshots. That makes me wonder if `Currency::round()` is causing this issue inside `getTotalPrice()`.

https://github.com/craftcms/commerce/blob/912e3167dfef4f7c084aa03ee60635b5a82d90cc/src/elements/Order.php#L1100

Recommend testing to see what the number is before and after it gets passed through that method.

N0ps32:
With the help of @ccchapman I was able to track down the bug.
The problem is Line 407 in the LineItem model:
https://github.com/craftcms/commerce/blob/912e3167dfef4f7c084aa03ee60635b5a82d90cc/src/models/LineItem.php#L407

salePrice is saved as a rounded value in the database before the tax adjuster gets a chance to produce a correct value. Rounding should be the last step not the first one.

If you change the line from 
`$this->salePrice = CurrencyHelper::round($this->saleAmount + $this->price);`
to
`$this->salePrice = $this->saleAmount + $this->price;`
everything works correctly.

price field in database: 12.4170
[...]

PR #697, opened 2019-02-08
Removed premature rounding that leads to errors

N0ps32:
Fixes issue #695 .
The values are rounded in the final total price calculation step anyway.
This makes sure the tax adjuster can work with the correct non-rounded value.

### DIFF
--- a/src/models/LineItem.php
+++ b/src/models/LineItem.php
@@ -403,8 +403,8 @@ class LineItem extends Model
         }
 
         // If a plugin used the above event and changed the price of the product or
-        // its saleAmount we need to ensure the salePrice works calculates correctly and is rounded
-        $this->salePrice = CurrencyHelper::round($this->saleAmount + $this->price);
+        // its saleAmount we need to ensure the salePrice works calculates correctly
+        $this->salePrice = $this->saleAmount + $this->price;
 
         // salePrice can not be negative
         $this->salePrice = max($this->salePrice, 0);


lukeholder:
This is a breaking change in behavior which other adjusters and plugins would not expect.

We could look at introducing a rounding policy setting that allows this to be configurable.

Every line item or adjustment amount has been rounded so that the total always makes sense. You are displaying the price of the line item in a currency format, or in 3 decimal places?

N0ps32:
@lukeholder I'm always displaying it with the currency filter with a precision of 2 decimals. The issue is that the value is wrong because the tax adjuster gets a rounded value instead of a precise one.

nfourtythree:
Just a weigh in here, a little. 

It would be great to have this as an option, over the years working on many UK/EU based e-commerce sites it has always yielded the most accurate results when waiting until the very last moment to do any rounding.

lukeholder:
@N0ps32 Are you only entering the price in 3 decimal places to you end up with a nicer looking price with your tax % added?

lukeholder:
I am a little confused, if the customer sees the 2 decimal rounded line item price, but the tax is calculated from the 3 decimal price, wouldn't that be confusing?

nfourtythree:
For example in the UK with VAT if you wanted an inclusive of VAT price of £9.99 you would need to enter an ex VAT price of £8.325.

8.325 * 1.2 (20% VAT) = £9.99

So the problem here is if you round the ex tax value earlier you get the wrong price

8.33 * 1.2 = 10.00 (9.996 before rounding). 

The way that this is normally dealt with, at least in my experience, is that for display purposes if you only ever want to show 2 decimal values to the end user you would do the following (very pseudo code):

```
Total inc tax = round( ex tax * tax )
9.99 = round( 8.325 * 1.2 )

Total ex tax = round( ex tax )
8.33 = round( 8.325 )

Tax = Total inc tax - total ex tax
1.66 = 9.99 - 8.33
```

This means all the numbers add together correctly in the listing and the tax man is happy. Obviously, I imagine some people do it differently but, so far, I haven't had any issues or notices from the government!

It also works the other way if you are entering your prices in the CMS inc tax. The calculations would follow:

```
Total inc tax = total
9.99 = 9.99

Total ex tax = round( ( total inc tax / 120 ) * 100 )
8.33 = round( ( 9.99 / 120 ) * 100 )

Tax = total inc tax - total ex tax
1.66 - 9.99 - 8.33
```

Hope this info at least helps in some way and isn't too confusing, there are definitely some oddities especially in the UK where there are provisions for "gaining and losing a penny".

N0ps32:
@nfourtythree is correct, this is also my issue.
  
I want my product to cost 14.90 so I priced it at 12.417. Adding 20% tax on top that's 14.90.  
However rounding it before the adjuster results in an incorrect price after tax.

lukeholder:
@nfourtythree @N0ps32 Does the customer ever see the price of the item in 3 decimal places?

If not, it sounds like the price in 3 decimal places is a work around to get nice looking price+tax totals. Maybe we can find a better solution to get nice looking (even) price+tax totals?

2 questions for now:

1) Is there ever a case where no tax will be applied to an item? i.e They supplied the VAT ID?
2) Why not use the included tax option to get nicer looking prices? This is what it was designed for.

Happy to find a solution that makes sense, and not ruling anything out.

nfourtythree:
Again, in my experience the answers to your questions are as follows:

1. Yes, there are tax-exempt products in the UK e.g kids clothing
2. For projects that I have been part of it has been simply due to a business decision, normally based on requirements that other systems the clients use e.g. stock control, offline sales etc

For me, it definitely makes sense storing using the inc tax price (as long as the numbers come out as above) and generally find that is the approach to take. Can be a sticking point with the client sometimes if it is different from what they are used to etc

N0ps32:
1. Yes we have tax exempt customers (businesses)

2. We have products in our system where we weren't sure about the tax rate beforehand (not everything is taxed 20% here). So we built the system on tax excluded prices for now. We also have a ton of custom logic that relies on prices that don't have the tax included. We'd have to rewrite a lot of code and do a ton of testing to change the whole shop to a different pricing model. We also need the price without tax on all Invoices we issue to comply with local tax laws.
  
I can see some advantages for having the tax included in the price however the state this feature is in is pretty misleading. The CMS happily accepts prices with more precision (one of the reasons we decided to go with this route in the first place), and then silently rounds them in the shop logic with no warning to the developer. 
  
Also this makes the feature that you can have taxes that aren't included in the price effectively useless since you can't rely on the correct prices being calculated in the end stage.

N0ps32:
@lukeholder any update on this issue?

fantasticmachine:
We have a similar issue. Consider this scenario:

VAT is 20%
1 Widget is priced at 1.41 ex VAT
1.41 * 1.2 = 1.692 

Because we need to show the price including VAT on the front end the customer is shown that the widget costs £1.69 each

A customer may purchase 100 widgets. We've found that the line item total is calculated as 1.692 * 100, which means the final line item total is shown as £169.20. Technically it's "correct" but from a UX perspective it's not great.

lukeholder:
We are team discussion internally tonight and hope to decide on the path forward on this issue soon. Thanks.

N0ps32:
@lukeholder any update after your discussion? This is still an issue for our client.

lukeholder:
Ok, we have discussed internally, and come to the following solution:

Rounding will be an option at either the item, line item, or order level. This should fix your issue, as you could select rounding at the order level if you want.

We will also introduce a rounding strategy option to round up, down etc.

Look out for the feature soon. Thanks for your patience while we did the research.

N0ps32:
@lukeholder do you know when we can expect to see this change?

stevenvandemoortele:
@lukeholder Any word on this? 
I'm also seeing incorrect rounded prices especially when using tax not included in price.

**Example**
price excl vat: €8.264 
tax rate: 21%

**Expected result**
price incl vat: €10.00 
```
8.264 * 1.21 = 9.99944 
rounded to 2 digits = €10,00
 ```

**Actual result**
price incl vat: €9.99
```
8.264 rounded to 2 digits = 8.26 
8.26 * 1.21 = 9.9946 
rounded to 2 digits = €9.99
 ```

lukeholder:
@stevenvandemoortele Will be adding rounded settings after the new 'order editing' features are in beta.

lukeholder:
Further discussion can be had on #695 as we are not using the PR and making our own solution.

Issue #715, opened 2019-02-14
Return order in JSON after payment made


ccchapman:
When paying for an order, it would be helpful to return the completed order for JSON responses.

https://github.com/craftcms/commerce/blob/2f5afe73bdd148b5b8661d88ee69521b943abfda/src/controllers/PaymentsController.php#L310

lukeholder:
Thanks fixed for the next release.

PR #716, opened 2019-02-14
Add order object to JSON response when paying for an order

ccchapman:
Closes https://github.com/craftcms/commerce/issues/715


lukeholder:
Thanks for the PR, this is fixed in 611d7cd304dc34ad328cf907a830b3f7ac3b2af1

Issue #747, opened 2019-02-23
uid is not included for Order Statuses


sjelfull:
### Description

As the title says.

When calling `getOrderStatuses()->getAllOrderStatuses()`, `_createOrderStatusesQuery()` doesn't include `uid`.

PR #748, opened 2019-02-23
Add uid to order status query

sjelfull:
Fixes craftcms/commerce#747


andris-sevcenko:
Thanks, man!

andris-sevcenko:
Wow. Bad case of Monday.

PR #781, opened 2019-03-11
Fix Product Type sources

boboldehampsink:
Product Type sources didn't support UID's yet so the id ended up in the project.yaml


andris-sevcenko:
You are the man! Thanks!

boboldehampsink:
Thanks @andris-sevcenko. Would you mind adding this to the changelog?

andris-sevcenko:
The real question is - why didn't you? :)

PR #786, opened 2019-03-13
Typo fixed

mikewink:
There was a "be" missing in the sentence.

`Only one purchasable can [be] ordered…`


andris-sevcenko:
Thanks!

PR #793, opened 2019-03-18
Fixed an issue where a product's metadata did not make it into Live Preview

boboldehampsink:
Product metadata (tax information, availableForPurchase, etc.) does not make it into livepreview.


boboldehampsink:
Bump

andris-sevcenko:
Thanks!

PR #804, opened 2019-03-27
Fixed active subnav item for commerce settings

janhenckens:
I noticed that the active sidebar item for most of the settings "new" or "edit" pages were not set correctly, this should fix that.


lukeholder:
Thankyou!

PR #818, opened 2019-04-07
Partial fix for Issue 814

magnessjo:
A [PR](https://github.com/craftcms/cms/pull/4107) has been entered in Craft for updating the info-hub modal.  

Note: I think adding the refund form to a modal would be better UX and make it consistent with the make a payment form.  Please advise is you would like me to make that change.


lukeholder:
Thanks!

PR #821, opened 2019-04-10
 Add before and after save customer events

cimocimocimo:
Adds two event triggers to Customers::saveCustomer() method. Allows a plugin to listen for the events and take action such as updating customer information in a payment gateway.


cimocimocimo:
Closing PR since it's not needed. Team let me know that the Customer records all extend Yii2 ActiveRecords and inherit all the events that I could ever want.

Issue #644, opened 2019-01-22
craft\commerce\models\OrderAdjustment::getSourceSnapshot() must be of the type array, null returned


N0ps32:
### Description

I updated my system from commerce 2 beta to commerce 2.0.1 and now my custom order adjustment stopped working. As soon as I add the order Adjustment I get this error:

```
TypeError: Return value of craft\commerce\models\OrderAdjustment::getSourceSnapshot() must be of the type array, null returned in /app/vendor/craftcms/commerce/src/models/OrderAdjustment.php:112
Stack trace:
#0 /app/vendor/yiisoft/yii2/base/Component.php(139): craft\commerce\models\OrderAdjustment->getSourceSnapshot()
#1 /app/vendor/craftcms/commerce/src/services/OrderAdjustments.php(144): yii\base\Component->__get('sourceSnapshot')
#2 /app/vendor/craftcms/commerce/src/elements/Order.php(1978): craft\commerce\services\OrderAdjustments->saveOrderAdjustment(Object(craft\commerce\models\OrderAdjustment), false)
#3 /app/vendor/craftcms/commerce/src/elements/Order.php(918): craft\commerce\elements\Order->_saveAdjustments()
#4 /app/vendor/craftcms/cms/src/services/Elements.php(515): craft\commerce\elements\Order->afterSave(false)
#5 /app/vendor/craftcms/commerce/src/controllers/CartController.php(304): craft\services\Elements->saveElement(Object(craft\commerce\elements\Order), false)
#6 /app/vendor/craftcms/commerce/src/controllers/CartController.php(287): craft\commerce\controllers\CartController->_returnCart()
#7 [internal function]: craft\commerce\controllers\CartController->actionUpdateCart()
#8 /app/vendor/yiisoft/yii2/base/InlineAction.php(57): call_user_func_array(Array, Array)
#9 /app/vendor/yiisoft/yii2/base/Controller.php(157): yii\base\InlineAction->runWithParams(Array)
#10 /app/vendor/craftcms/cms/src/web/Controller.php(109): yii\base\Controller->runAction('update-cart', Array)
#11 /app/vendor/yiisoft/yii2/base/Module.php(528): craft\web\Controller->runAction('update-cart', Array)
#12 /app/vendor/craftcms/cms/src/web/Application.php(297): yii\base\Module->runAction('commerce/cart/u...', Array)
[...]

PR #823, opened 2019-04-11
Issue 644 - Default the $_sourceSnapshot to an array

harnasz:
When bumping our Craft Commerce version up we had the following error:

```
craft\commerce\models\OrderAdjustment::getSourceSnapshot() must be of the type array, null returned
```

This was originally reported here:

https://github.com/craftcms/commerce/issues/644

If people are bumping their versions and use adjusters this PR will save them on having to amend their adjusters like so:

```
$discountAdjuster->sourceSnapshot = [];
```

By default `$_sourceSnapshot` to an array.

### DIFF
--- a/src/models/OrderAdjustment.php
+++ b/src/models/OrderAdjustment.php
@@ -58,7 +58,7 @@ class OrderAdjustment extends Model
     /**
      * @var mixed Adjuster options
      */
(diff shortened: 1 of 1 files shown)


lukeholder:
Thanks!

PR #831, opened 2019-04-22
Use uid instead of id for redactor sources

davist11:
The Redactor **Link to a product** and **Link to a variant** options were returning no results since it was using the `id` instead of `uid`


andris-sevcenko:
Thanks!

PR #832, opened 2019-04-23
Switch the Order element to override `Element::tableAttributeHtml()`.

joshangell:
This PR changes the Order element to override `Element::tableAttributeHtml()` instead of `Element::getTableAttributeHtml()` and fixes an issue where plugins and modules couldn’t add custom index table attributes via the `Element::EVENT_SET_TABLE_ATTRIBUTE_HTML` event.

That event gets fired in `Element::getTableAttributeHtml()` and the method in the Order element class wasn’t then falling back to using that in the default part of the switch statement (line 1820).

This change is in line with the method used in the Entry element class.


andris-sevcenko:
Thanks!

joshangell:
Thank you kind sir!

PR #834, opened 2019-04-24
Fix default norwegian translations from no to nb.

eddiedale:
Should be the same as used in Craft otherwise for Norway (nb).


nfourtythree:
Hey @eddiedale 

Thanks for this, this was fixed in 77d8a3ce66fae12c841986cc581dba7d5cc2c252

Thanks.

PR #836, opened 2019-04-26
Move the 'stopProcessing' check, fix bug with multiple sales

nilsenpaul:
I ran into a bug where, if two sales where applied, of which the second one was only applied to purchasable A, it would not be applied if the first sale had 'stopProcessing' set to true. This change will make sure processing is only stopped if the first sale *applies* to the purchasable, not only if the sale exists and has stopProcessing set to true.


lukeholder:
Thanks!

PR #855, opened 2019-05-10
Line item quantity validation fix

VitaliiTsilnyk:
Fixes a couple of problems with order line item quantity validation.

First of all, the `$qty` variable is being calculated only one time per LineItem object instance because Yii creates all validators when you request them the first time and then reuses them inside the same model object for all subsequent validations. And although the quantity validation functions will be called each time the validation runs, the actual quantities of purchasables will not be updated, they will be stored in the closure and you will always validate the first set of purchasable quantities each time you run a validation. I understand that this calculation had been put outside of the validator to optimize and minimize the code, but it will fail in case when you:
1. Validate the order with correct quantity.
2. Change the quantity of any line item.
3. Re-validate the order. And no matter what quantity you have in your LineItem object right now, validation will always pass, as long as it passed for the first time.

Also, I believe the current version of the code contains a typo:
```php
// count new line items
if ($lineItem->id === null) {
```
I think there should be the `$item` variable used instead of `$lineItem`. Because in case the current line item is new, only it's quantity will be used, ignoring all other line items in the order with the same purchasableId.

My fix encloses the quantity calculation into a closure, allowing to re-calculate the quantities each time properly while keeping the code minimal. And although I had to get rid of the optimization, I don't think it will have any noticeable impact on the performance.

### DIFF
--- a/src/elements/Variant.php
+++ b/src/elements/Variant.php
@@ -600,29 +600,17 @@ class Variant extends Purchasable
             return [];
         }
 
-        $qty = [];
-        foreach ($order->getLineItems() as $item) {
-            if (!isset($qty[$item->purchasableId])) {
-                $qty[$item->purchasableId] = 0;
-            }
-
-            // count new line items
-            if ($lineItem->id === null) {
-                $qty[$item->purchasableId] = $lineItem->qty;
-            } else {
-
-                if ($item->id == $lineItem->id) {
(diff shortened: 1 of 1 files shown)


lukeholder:
This is a great PR, thanks for the detailed explanation also. We appreciate it! Merging shortly.

lukeholder:
Changelog item: efe51a8cc02e5802665a40b05c794c05188b0584

PR #857, opened 2019-05-10
Fix isUnpaid condition in OrderQuery::beforePrepare()

martinleveille:
This fix the order query to get unpaid orders. This bug was introduced in the 2.1.5.2 release.


andris-sevcenko:
Thank you!

PR #863, opened 2019-05-14
Add missing argument to getPaymentFormHtml() example

medoingthings:
Without the empty object as an argument it throws a "Too few arguments" error


andris-sevcenko:
Much obliged!

PR #872, opened 2019-05-17
Improve IDE return for Variants service

mildlygeeky:
Rather than promising to return an ElementInterface, instead show that a Variant (or null) will be returned (as the Products service already does, though for Products)


lukeholder:
Thanks

PR #887, opened 2019-05-30
Fix: make email null if empty on a order

mediabeastnz:
I noticed when an order is created and an email is not yet set it doesn't default to NULL like most of the other fields on the order.

This should just make it a bit more consistent especially when you need to query for orders that have emails set.


lukeholder:
Thanks

Issue #892, opened 2019-06-06
Order search:  Impossible to search by full name


echantigny:
### Description
Because the first name and last name are index separately, it's impossible to do a search on a full Shipping or billing name.  Because these 2 fields are searched together very often, I would suggest getting rid of the separate fields and merge first and last in 1 field, like it is done for users fullname.

![2019-06-06 14_06_26-localhost _ MySQL _ greatboardgames_live _ searchindex _ phpMyAdmin 4 7 9](https://user-images.githubusercontent.com/4933333/59055804-b79ae680-8864-11e9-964e-2f2118a61451.png)
  

### Steps to reproduce

1.
2.

### Additional info

- Craft version: Craft Pro 3.1.29
- Craft version: Craft Pro 3.1.29
- PHP version: 7.2.4
- Database driver & version: MySQL 5.7.21

echantigny:
PR https://github.com/craftcms/commerce/pull/893

echantigny:
This fix only works for resaved items after the fact.  Any chance there could be a way to resave all the order indexes ?

lukeholder:
Thanks, updated for the next release.

PR #893, opened 2019-06-06
Fix for issue #892

echantigny:
I did not remove the first and last name from the index in case it is necessary somewhere else,  but if not, they could be removed.

Also, this only works for resaved items after the fact.  Any chance there could be a way to resave all the order indexes ?


echantigny:
Do not merge this one.  It broke the Add To Cart action.  I'll send a new PR.

PR #895, opened 2019-06-06
Fix for the Order Full Name search 

echantigny:
https://github.com/craftcms/commerce/issues/892


lukeholder:
This has been resolved in 2.1.8

PR #904, opened 2019-06-13
Update example code to include hashed `cancelUrl` value

sjcallender:


bossanova808:
Would be nice if there was a `{{ cancelUrlInput() }}` to match the others actually..

PR #921, opened 2019-06-26
Boolean fixes for Postgres

boboldehampsink:
Fixes errors like these on Postgres:

```
Undefined function: 7 ERROR:  operator does not exist: boolean = integer
LINE 3: WHERE "default"=1
                       ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
The SQL being executed was: SELECT "shippingCategories"."id", "shippingCategories"."name", "shippingCategories"."handle", "shippingCategories"."description", "shippingCategories"."default"
FROM "commerce_shippingcategories" "shippingCategories"
WHERE "default"=1
LIMIT 1  
```

### DIFF
--- a/src/services/OrderStatuses.php
+++ b/src/services/OrderStatuses.php
@@ -145,7 +145,7 @@ class OrderStatuses extends Component
         }
 
         $result = $this->_createOrderStatusesQuery()
-            ->where(['default' => 1])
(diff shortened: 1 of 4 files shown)


lukeholder:
Thanks1

PR #936, opened 2019-07-21
Discounts - Add template hook and `beforeSave`, `afterSave` and `afterDelete` events

engram-design:
We have a need for some custom logic with discounts, but would like to extend the functionality of discounts rather than completely roll our own solution or duplicate code. For example, we want to implement applying discounts to chosen product types.

This PR includes a template hook to allow additional fields to be added to the first tab of discounts. Then I've also added events on the save and delete of a discount, so through a custom module, we can listen on these events, and store our custom data in a separate table, for applying later.

### DIFF
--- a/src/events/DiscountEvent.php
+++ b/src/events/DiscountEvent.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * @link https://craftcms.com/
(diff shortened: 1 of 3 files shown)


engram-design:
You're my favourite Commerce developer @lukeholder 🎉

PR #941, opened 2019-07-27
Update CRAFT_VENDOR_PATH

putyourlightson:
Assuming the plugin is installed in `vendor/craftcms/commerce`, the `_bootstrap.php` file will be in `vendor/craftcms/commerce/tests` and therefore vendor path is 3 directories up, therefore:

```
define('CRAFT_VENDOR_PATH', dirname(__DIR__, 3));
```


lukeholder:
@putyourlightson If you are running tests from your project, it will have its own _bootstrap file right? This bootstrap is for running tests from the commerce repo.

Maybe I am missing something @gtettelaar ?

gtettelaar:
@lukeholder Essentially yes. Not entirely sure if @putyourlightson was running into something that needs this change but essentially this config setting is set up to allow the running of commerce tests through CI e.t.c...

putyourlightson:
Ok I understand now. I was trying to run the commerce plugin tests with the plugin installed locally and running into the vendor path issue based on my setup. Thanks for the responses!

Issue #952, opened 2019-08-06
Add refunded item back to available quantity


DomLip94:
### Description

As far as I'm aware, refunding an order doesn't increase the available quantity by re-adding that quantity back to the product.

If this is a bug please let me know. 

I am happy to make the changes myself and submit a PR if you wish.

Cheers

lukeholder:
This for the PR, does this take into account partial refunds? Partial refunds is possible in the current refund feature.

lukeholder:
You are right, we don't have a current restock ability when refunding.

DomLip94:
Unfortunately not at the minute, I do intend to add it down the line though. Although it probably means a re-write of some functionality to avoid second guessing what's actually being refunded

DomLip94:
If it is easier I can just create a plugin to release. I'd need to have a serious look at the code before starting something like this

lukeholder:
Closed in favor of #1406

PR #955, opened 2019-08-07
FR #952: Add refunded item back to available quantity

DomLip94:
Basic shop functionality?

FR here https://github.com/craftcms/commerce/issues/952


andris-sevcenko:
Another thing to consider is that this makes the assumption that a refund means a returned product, which is not always the case. What if two items were refunded of which one was broken on delivery but the other one was returned?

DomLip94:
I'm gonna go back to the drawing board with this one. All valid points. Will come back soon

lukeholder:
Thanks for the PR @DomLip94! We look forward to future contributions.

PR #966, opened 2019-08-21
Fixes #592 saving discount percentage with comma decimal separator

### LINKED ISSUE #592: Discount percentages and comma's
### Description
Our editors use comma's as a decimal separator. That works fine, except when they try to add a discount with decimals. Then 7,5% becomes 7% after saving. If you use a dot as the decimal separator, the correct discount percentage gets saved. However, if you re-open that discount, the field says `7,50%`. Saving that sets the percentage back to 7%.

### Steps to reproduce

1. Use a language with commas as decimal separator (Dutch, in our case)
2. Whilst using Dutch, try to navigate to discounts and create a new one
3. Use 7,5% as the discount percentage and save
4. Reopen that discount. Percentage is now 7%.

nfourtythree:


lukeholder:
@nfourtythree Will wait for the other number field fixed (we discussed today) be added to this branch before merging.

nfourtythree:
@lukeholder have pushed up some more commits for this. As well as a migration for the `purchaseTotal` column.

Let me know if it is all good and I will sort the `CHANGELOG` conflicts

lukeholder:
Should be good to merge. Just made a couple of cleanups, and fixed the changelog.

src/migrations/m190821_150226_discount_purchaseTotal_update.php
```diff
@@ -16,7 +16,6 @@ class m190821_150226_discount_purchaseTotal_update extends Migration
     public function safeUp()
     {
         $this->alterColumn('{{%commerce_discounts}}', 'purchaseTotal', $this->decimal(14, 4));
-        true;
```
nfourtythree: @lukeholder doesn't the `safeUp` method have to `return true` for a successful migration?

PR #970, opened 2019-08-22
Add `EVENT_MODIFY_SHIPPING_PRICE_FOR_ORDER`

engram-design:
Would love to get some feedback on this, whether it could be done better, another way, or if a valid approach.

We would like to modify the amount of a particular shipping method/rule depending on a set of circumstances. Its more advanced than this PR covers, but basically adds more logic to discounts, allowing you to set the price of certain shipping methods with a coupon code.

Adding this event allows us to modify the shipping price for a particular order. Not sure if you'd like to name the event differently as well.

I am aware of the `Remove all shipping costs from the order` setting, but that's not desirable, as:
- We only want certain shipping methods affected
- We want to set the price to something specific, not just free

### DIFF
--- a/src/base/ShippingMethod.php
+++ b/src/base/ShippingMethod.php
@@ -11,6 +11,7 @@ use Craft;
 use craft\base\Model;
 use craft\commerce\elements\Order;
 use craft\commerce\errors\NotImplementedException;
+use craft\commerce\events\ModifyShippingPriceEvent;
 
 /**
  * Base ShippingMethod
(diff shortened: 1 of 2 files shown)


lukeholder:
`getPriceForOrder()` is not used for applying cost to the cart, it duplicates logic in the shipping adjuster, including shipping costs added to line items as well as the base shipping cost. If someone used this event, it wouldn't affect the price of the order. We would need to tackle the adjuster side as well. 

The cleanest solution would be to write you own shipping method wouldn't it?

engram-design:
@lukeholder Ah, I see. I think I'll go the route of a custom shipping method then (which is otherwise, just a standard method)

engram-design:
The only real issue with a custom shipping method, is you loose the CP configuration side of things. Unless I'm missing something where you could override them?

nfourtythree:
@engram-design could you not keep the Shipping method as it is at the moment and just write and adjuster to do the extra bit. 

You could check if the shipping method is matching by using the `matchOrder` method in the `ShippingRuleInterface` and then apply the adjuster accordingly. In your instance, it would be a minus amount (I assume?).

Let me know if this makes any sense, or if you need to discuss it any further.

engram-design:
I should've mentioned that's what I ended up doing, so thanks!

PR #973, opened 2019-08-23
Fix for sites that require shipping method at checkout

dhahn:
I updated our commerce version to 2.1.12 and was unable to checkout on our local instance. Traced the issue to what I think was a type on a property name: 

Update property `getShippingMethodId` to method call `getShippingMethodId()`


nfourtythree:
Good spot, @dhahn thanks for the PR!

brandonkelly:
2.1.12.1 is out now with this. Thanks again!

dhahn:
Wow awesome work. Thanks y'all!

PR #974, opened 2019-08-23
Basic Velocity Filter

jmauzyk:
The payment gateway we use with Commerce unfortunately doesn't have any velocity filter. This is my attempt at a basic configurable velocity filter to limit the number of transaction attempts by IP. Would it be possible to implement this or something similar to help prevent excessive card testing by fraudsters?

### DIFF
--- a/src/models/Settings.php
+++ b/src/models/Settings.php
@@ -170,6 +170,26 @@ class Settings extends Model
      */
     public $gatewaySettings = [];
 
+    /**
+     * @var bool
+     */
+    public $enableVelocityFilter = false;
+
+    /**
+     * @var mixed
+     */
+    public $velocityFilterWindowDuration = 3600;
+
+    /**
+     * @var int
+     */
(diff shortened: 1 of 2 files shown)


nfourtythree:
Hey @jmauzyk 

Thanks for the PR, an interesting concept and something we can discuss at this end.

It is a tough one, we would have to make sure that the IP is tied to a specific order. Using IP as the determining factor is a difficult one. Due to the fact, things like mobile network carriers can often use the same IP for many devices so you wouldn't want to block a lot of people based on one user's failed attempts.

A lot of gateway providers handle this at their end so can't guarantee that this will be put in, but definitely, something for us to think about.

Thanks.

nfourtythree:
Hey @jmauzyk 

Going to close this for now as generally, this is something that the payment gateways handle. Are you having a lot of issues with this at the moment?

I have created a ticket for an enhancement for the future that will look at introducing an attempted payments threshold on a order #982

PR #975, opened 2019-08-26
Added parsing of environment settings in emails

thejoshsmith:
+ fromEmail & fromName are now parsed from environment variables.


nfourtythree:
Hey @thejoshsmith 

Thanks for the PR, there are a few other bits that need to happen with this before getting merged in. Good spot though!

PR #980, opened 2019-08-27
Order export doesn't use paid status filtering

engram-design:
When trying to export Unpaid or Paid orders, using the sidebar filtering for these, orders aren't correctly filtered and exported. This is because there's a lack of checking if we're filtering by order statuses, or paid status. This PR includes that.

As an aside, there's also probably unnecessary order status fetching via https://github.com/craftcms/commerce/blob/develop/src/services/Reports.php#L95, since the parameter to the function is already an ID, and that's all that's used. Thinking it might just be better to pass the order status from the controller in a single call (which is being done anyway - https://github.com/craftcms/commerce/blob/develop/src/controllers/DownloadsController.php#L92)

### DIFF
--- a/src/controllers/DownloadsController.php
+++ b/src/controllers/DownloadsController.php
@@ -9,6 +9,7 @@ namespace craft\commerce\controllers;
 
 use Craft;
 use craft\commerce\Plugin;
+use craft\commerce\elements\Order;
 use HttpInvalidParamException;
 use Throwable;
 use yii\base\Exception;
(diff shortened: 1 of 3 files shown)


lukeholder:
@engram-design How are you filtering paid and unpaid at the moment from sources? We don't have that built in at the moment.

engram-design:
@lukeholder I **knew** I should've checked that first, but for some reason thought it was stock behaviour. You're right, I've customised stuff on the order sources to show these options.

Will look at another way to go about this.

PR #981, opened 2019-08-27
Added details and example for commerce.php file

andrewcgregory:
Added add additional details and an example of adjusting these settings using environment config arrays. Copied the same format as https://docs.craftcms.com/v3/config/config-settings.html to be consistent.


lukeholder:
Thanks!

PR #999, opened 2019-09-09
Check for COMMERCE_PAYMENT_CURRENCY const, and set paymentCurrency

bymayo:
I did a Craft 2 > Craft 3 upgrade where the `paymentCurrency` was set with a const in the `index.php` file previously,  due to it being multi site.

In the docs (https://docs.craftcms.com/commerce/v2/payment-currencies.html#switching-currencies) it says you can set the `COMMERCE_PAYMENT_CURRENCY` const still, but this doesn't work in Craft 3 / Commerce 2.

Doing a bit of digging, it seems this const is never checked and set for payment currency. Adding this code back in works perfectly still for anyone using sites/locales.

(I opened a question about this also - https://craftcms.stackexchange.com/questions/31442/craft-3-commerce-2-multisite-commerce-payment-currency)

If this has been left out due to a better way of doing this, please let me know (And update the docs)


engram-design:
Likewise interested to know how to achieve this. Not a massive fan of setting every call to update the cart with the correct currency.

In our case, we have a multisite setup with a currency per-site. There is no way to switch currencies on each site, as we don't want that. You order on the NZ site, and everything should be in NZD for instance.

We're also not using the `CRAFT_LOCALE`, and have a single index.php file for all our multi-sites. So even if this was put back in, we'd have to conditionally set this constant depending on the site currently being viewed.

What's the accepted way to do this?

bymayo:
@engram-design Your setup does sound similar to ours. The client requested that the AUS site be in AUD for instance, with no ability to change the currency.

I'm just still wondering why this feature would be dropped in Commerce 2, when it was available in Commerce 1 when locales weren't even a thing. With no solution to actually implementing this - Especially with muti-site/locales being a major change (addition) to Craft 3.

I'm in a desperate need to have this PR merged, or have a solution for this as we're currently about 10 builds behind due to us using a forked version of Commerce to add this feature.

Maybe @lukeholder could shed some light on this.

iamtompickering:
@lukeholder just opening this up again as I have tried to add `define('COMMERCE_PAYMENT_CURRENCY', 'USD');` into one of my site locales `index.php` files but it doesn't seem to work?

Just a note here though, it does seem to work in the main `index.php`, but it doesn't pick up the ones set in the other sites `index.php` files.

PR #1009, opened 2019-09-17
Product Fixture

boboldehampsink:
Craft 3.2 comes with fixtures for its element types. This is the fixture for the product element type.


lukeholder:
Thanks will take a look at this next week

boboldehampsink:
cc @gtettelaar

PR #1011, opened 2019-09-19
fix: update order pdf url to use cpUrl rather than siteUrl

jamesirving:
When using craft in a headless setup, if the admin user clicks 'Download PDF' on an order they get a 404 in a new tab as it is directing the user to the siteUrl rather than the cpUrl. 

Changing the URL to the cpUrl resolved this issue and triggered the pdf download as expected.


jamesirving:
@lukeholder would be good to get your thoughts on this.

lukeholder:
Fixed in 23d0d9b7b61d1f756a125a79232538c85a54f0ba

MichaelWheldon:
@lukeholder hey just checking on when this was getting released?

PR #1017, opened 2019-09-24
Check the order->dateOrdered is object when completing order

samuelbirch:
It was throwing an error when the order->dateOrdered was being returned as a DateTime object and then trying to use it to create a new DateTime object. So this is a little check before trying to create a new object.


nfourtythree:
Hi @samuelbirch 

Thanks for the PR. In what scenario was the error manifesting?

Thanks.

samuelbirch:
Hi @nfourtythree 

If an order has been created but not completed and no transactions (guess you could use an active cart, we created the order with admin-orders), then if you click the "complete order" button, at that point the error was being thrown.

nfourtythree:
Ah ok, so this is because an order shouldn't have a `dateOrdered` value before it is completed if that makes sense. Completing the order is what gives the order its `dateOrdered` value.

So in the normal operations of Commerce this error wouldn't exist.

samuelbirch:
If i add something to my cart as normal via the frontend, the login to the backend and go to active carts. If i view the details of that active cart and click the "complete order" button I get the error. Admin orders plugin not involved in this case.

nfourtythree:
Hmm, I thought I had just tested this and it was working fine. Let me go back and double-check.

Thanks for the info!

nfourtythree:
Hey @samuelbirch 

Have just tested it again and I did the same. Create a cart in the front end all the way up to payment and then completed it in the CP and it worked correctly.

Here is a video: http://x.n43.me/20a61713d4c1

It seems like there might be something else at play here?

samuelbirch:
yeah it seems like there could be something else at play. 

Ok, leave it with me. I'd do some more testing to try and pin point the source.

Thanks for taking the time with this :)

lukeholder:
Please let us know if you figure it out. Thanks.

PR #1025, opened 2019-10-03
fixed a typo

hiasl:
for select gateways -> for selected gateways


nfourtythree:
Thanks for the PR. This is technically already a correct sentence so will just leave it as is for the moment.

PR #1033, opened 2019-10-06
Tip to install plugin in Craft if using Composer.

SiCoUK:
Just added a little note to make sure it's clear that you actually have to install the plugin in Craft if you use Composer.

As a new comer wanting to use Commerce without having disabled with Craft CMS Plugins it was not immediately obvious I had to actually install the Plugin. In hindsight I should of used the plugin store!


lukeholder:
Thanks

PR #1034, opened 2019-10-08
Cleanup ProductFixtures

gtettelaar:
Cleans up #1009. 

`deleteElement()` was not necessary because commerce already deletes all variant's in the `ProductElement::afterSave()`.


lukeholder:
Thanks Giel!

PR #1035, opened 2019-10-08
Link to correct twig function

AlexJenter:
Possibly wrong, but it think the function docs musst have moved from `https://docs.craftcms.com/v3/dev/functions.html#seq-name-length-next` to `https://docs.craftcms.com/v3/dev/functions.html#seq` at some point in time


nfourtythree:
Thanks @AlexJenter !

PR #1050, opened 2019-10-14
Adds line item options management to the new order edit view

joshangell:
This is an attempt to add the ability for better line item options management whilst creating or editing new line items in the new order edit view.

The options form is configurable via the commerce config file in the following format:

```php
return [
    'lineItemOptionsConfig' => [

        // Store config blocks under each purchasable’s class type
        \craft\commerce\elements\Variant::class => [
            'plaintext_key' => [
                'label' => 'Plain text option', // Required
                'type' => 'text', // Required
                'required' => true // Optional
            ],
            'dropdown_key' => [
                'label' => 'Dropdown option',
                'type' => 'dropdown',
                'required' => true,

                // Required for dropdown type
                'options' => [
                    [
                        'label' => 'Option One',
                        'value' => 'option_one'
                    ],
                    [
                        'label' => 'Option Two',
                        'value' => 'option_two'
                    ],
                    [
                        'label' => 'Option Three',
                        'value' => 'option_three',

                        // Optional
                        'default' => true
                    ]
                ]
            ],
            'number_key' => [
                'label' => 'An option that should be a number',
                'type' => 'number',
                'required' => true
            ],
        ],
        \verbb\events\elements\Ticket::class => [
            'plaintext_key_one' => [
                'label' => 'Plain text option one',
                'type' => 'text',
                'required' => true
            ],
            'plaintext_key_two' => [
                'label' => 'Plain text option two',
                'type' => 'text',
            ],
            'plaintext_key_three' => [
                'label' => 'Plain text option three',
                'type' => 'text',

                // Optional
                'default' => 'Some default text'
            ],
            'number_key_one' => [
                'label' => 'Optional number',
                'type' => 'number',
            ],
        ]
    ],
];
```

If a config block is present for a purchasable type then the options form will appear, if not it falls back to the existing prism editor.

The form also validates and spits back error messages to the parent component in a custom `validated` event.

There is currently an issue when the form initially loads in that it runs the validation too early, but I’m submitting this now in case its not the direction you want to take. If it is the sort of thing you want to have in the commerce core then I’d be happy to continue working on it.

---

### When adding a new line item
image

### Validation error
image

### Editing an existing line item
image


lukeholder:
Exactly what we were looking to do, thanks!

Couple of notes:

- We should look to use a common form json definition format (like json-schema but for forms)
- We should look to extend the definitions to be provided from more than just settings (allow PHP events to provide the definitions)
- We should have the definition keys be more fine-grained that just the Purchasable class. e.g allow productType ID's in the key. (Like Element sources)

lukeholder:
Will make a new PR for the new feature branch where we will work further on this.

joshangell:
@lukeholder Brilliant, thanks for this Luke - looking forward to it landing!

PR #1052, opened 2019-10-14
Clean up addition of X-Powered-By header

putyourlightson:
This PR refactors some code that feels a big legacy and that can be achieved in a single line of code. It also feels better to me to have a single `X-Powered-By` header per plugin (as SEOmatic and Blitz do), rather than comma separating them.

Old header: 
```
X-Powered-By: Craft CMS,Craft Commerce
```

New headers: 
```
X-Powered-By: Craft CMS
X-Powered-By: Craft Commerce
```

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -467,11 +467,7 @@ class Plugin extends BasePlugin
             $headers = Craft::$app->getResponse()->getHeaders();
             // Send the X-Powered-By header?
             if (Craft::$app->getConfig()->getGeneral()->sendPoweredByHeader) {
-                $original = $headers->get('X-Powered-By');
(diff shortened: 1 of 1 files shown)


lukeholder:
Thanks Ben

3rd parties (like wappalyzer) don't pick up additional `X-Powered-By` headers, which is why we modify the initial header, rather than adding a new one.

Maybe seomatic and blitz should use our same implementation.

putyourlightson:
This was fixed in Wappalyzer in 2017 (https://github.com/AliasIO/Wappalyzer/issues/1213), although  I don't know about other similar applications. If you're saying that this is the _official_ way to do it in Craft then I agree that plugin developers (myself included) should follow suit.

PR #1055, opened 2019-10-17
Return orderlines by dateCreated from LineItems service

thejoshsmith:
Order Lines retrieved using the LineItems service `getAllLineItemsByOrderId` method are now returned in order of creation date.

Use case is one of our clients added products to their order in a specific way, and is querying why they appear in a different order in the CMS.

### DIFF
--- a/src/services/LineItems.php
+++ b/src/services/LineItems.php
@@ -99,6 +99,7 @@ class LineItems extends Component
         if (!isset($this->_lineItemsByOrderId[$orderId])) {
             $results = $this->_createLineItemQuery()
                 ->where(['orderId' => $orderId])
+                ->orderBy('dateCreated')
                 ->all();
 
             $this->_lineItemsByOrderId[$orderId] = [];


nfourtythree:
hey @thejoshsmith 

Thanks for the PR, we needed to do a few extra steps to get this working as required so have created a new PR at #1058 which you can follow.

engram-design:
It'd be great if this was configurable, as I would've thought line items be ordered by ID - their sequence in being created.

We have an import script set to import line items for large orders, and this causes the dates to be slight off (no idea why, but here we are). Each time we import the order changes, despite PHP being a synchronous.

We can indeed order by IDs, which I would've though the default?

`{% set lineItems = order.lineItems | sort((a,  b) => a.id <=> b.id)  %}`

We've even tried `sleep()` between our `saveLineItem($lineItem)` calls, but there's no difference. In fact, that won't help, because the latest-dated line item will always be first.

PR #1058, opened 2019-10-18
Order lineItems by dateCreated by default

nfourtythree:


nfourtythree:
In the cart, lineItems are now ordered by 'dateCrated DESC' by default. They also keep their initial `dateCreated`.

You can order them by a different sort order if required. For example to order them by oldest to newest you could do the following:

```twig
{% set lineItems = cart.lineItems|sort((a,  b) => a.dateCreated <=> b.dateCreated)  %}
{% for item in lineItems %}
…
{% endfor %}
```

PR #1065, opened 2019-10-22
Move sending email to queue

engram-design:
Tested pretty thoroughly, but the tricky part here is how to handle serialising the order data. As we can't guarantee immediate processing of this queue, we need to make sure we use the 'snapshot' of the order at the time.

This is sort of achieved through `$order->setAttributes($this->order);` where `$this->order` is an array of the order data. I'm ditching the lineItem and adjustment - probably not ideal? I feel like they won't change all that much?

We also need create objects so we don't have to refactor things, or change emails.

### DIFF
--- a/src/models/Settings.php
+++ b/src/models/Settings.php
@@ -172,6 +172,11 @@ class Settings extends Model
      * @var bool
      */
     public $pdfAllowRemoteImages = false;
+    
+    /**
+     * @var bool Whether the send all emails using queue jobs
+     */
+    public $useQueueForEmails = false;
 
     /**
      * @var string The order reference format
--- a/src/queue/jobs/SendEmail.php
+++ b/src/queue/jobs/SendEmail.php
@@ -1,0 +1,61 @@
+<?php
+namespace craft\commerce\queue\jobs;
+
+use craft\commerce\Plugin;
+use craft\commerce\elements\Order;
+use craft\commerce\models\Email;
+use craft\commerce\models\OrderHistory;
+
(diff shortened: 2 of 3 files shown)


Mosnar:
Just my two-cents, but it might be a good idea to add this behavior as optional, controlled by the config and defaulted to off. Not all environments are going to have access to async queues, and having order emails being sent out only when someone logs into the CP could be an issue - especially if you have a long queue.

engram-design:
@Mosnar Yep, probably not a bad suggestion!

engram-design:
@Mosnar FYI, have updated to opt-out of this behaviour. @lukeholder @nfourtythree  This is not a non-breaking change now, which I feel a bit better about. This has drastically improved our checkout completion time.

On our local testing, sending 2 emails, one with a PDF took about 9s. With this fix in place, its down to 2s. I'm sure this would be faster on a production server.

gojira32:
+1 on this. Our checkout is super slow right now and we're getting complaints from customers who are sitting there for 5-7+ seconds and hitting the back button because "our website doesn't work".

I'm guessing for every 1 person that does bother to get in touch with us about it, there are 10+ who just give up.

keyurshah:
moving the email processing to queues is definitely needed for the checkout process.  looks amazing!

lukeholder:
Going to make some adjustments, but will be merging this into the next release. thanks!

engram-design:
@lukeholder awesome, let me know if you need a hand!

PR #1072, opened 2019-10-24
Fix PDF not rendering

Mosnar:
I was trying out the 2.2 release and got an error when trying to download an order PDF due to `templatePath` not existing on the PdfEvent.


lukeholder:
Thanks!

PR #1091, opened 2019-10-29
Remove references to ModifyCartArray

oddnavy:
Small correction to docs for modify cart info event.


nfourtythree:
Thanks for this!

PR #1093, opened 2019-10-30
Fix eagerloading map for Variant

dhahn:
We're getting an error that `commerce_variants` doesn't exist when trying to eagerload the product. Tracked it down to this line. This fixed it for us locally, but would be sick to have merged in so we can avoid doing 2 eager loading statements for variants and products.


lukeholder:
Thanks!

dhahn:
Wow speedy! Thanks!

PR #1120, opened 2019-11-12
Fix a type hinting issue in LineItem model

abryrath:
If options are not explicitly set, calling $lineItem->getOptions() throws an error due to its return type (array).


nfourtythree:
This has been fixed in e89130607888b63301a2a6ef1b1e3b89b77a3aa5 and will be in the next release.

Thanks!

PR #1130, opened 2019-11-14
Remove Customer fieldtype

nfourtythree:


nfourtythree:
@lukeholder I noticed there is a reference to `Customer::class` in a previous migration.

Will that migration break now that that class has been removed? See: https://github.com/craftcms/commerce/blob/3.0/src/migrations/m160531_154500_craft3_upgrade.php#L37

lukeholder:
@nfourtythree it will if the class has been removed. Switch that migration from `Customer::class` to the class string.

PR #1135, opened 2019-11-18
Fix deprecation error

billythekid:


bossanova808:
Would be handy to get this one in, for those of us that your hard exceptions for deprecation errors in dev environments....

nfourtythree:
Hey @billythekid and @bossanova808 

Unfortunately, I can't accept this PR. This would be a breaking change for many Commerce users which is the reason for the deprecation log. This update has been made in `3.0` (ef17c7010d043e3bda8b69aea340f602fe33a0d6).

However, what I have done is pushed and update (c426e4c049b49db4da948b2d5c817646fd142686) so the log will only be fired for those people that are not using the `fullName` property and relying on the computed behaviour.

Thanks!

PR #1139, opened 2019-11-20
Update _input.html

fantasticmachine:
Forces orders to be listed by dateOrdered.
Also made a change to how status is displayed to accommodate fetching orders in this way.
And limited to only completed orders.

Note: Would be "better" to use Twig's sort filter on the customer.orders array but this causes a recursion error.

### DIFF
--- a/src/templates/_fieldtypes/customer/_input.html
+++ b/src/templates/_fieldtypes/customer/_input.html
@@ -1,5 +1,7 @@
 {% if customer %}
-    {% set orders = customer.orders %}
+    {# sorts the returned orders by date ordered #}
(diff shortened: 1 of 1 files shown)


nfourtythree:
Hi @fantasticmachine 

Thanks for the pull request.

I am going to close this as we have deprecated the Customer field in favour of a specific Customer Info Tab. The field will be removed in the upcoming version 3.

We are also looking at making adjustments to the data in the tab for version 3.

Thanks!

fantasticmachine:
That’s great, thanks!

I’m only submitting it because I need to do some customisation to the field and that seemed like an easy thing to sort out.

I look forward to seeing the new version.

PR #1152, opened 2019-11-29
Fix some strings translatable in product edit page

watarutmnh:


nfourtythree:
Hey @watarutmnh thanks for this!

PR #1154, opened 2019-11-29
Fix target translation file in product index page

watarutmnh:
Fix to use commerce.php instead of app.php for translation.


nfourtythree:
Hey @watarutmnh thanks for this!

PR #1159, opened 2019-12-02
Update Order.php

fantasticmachine:
Ability to show Coupon Code within the order element index


nfourtythree:
Hey @fantasticmachine 

I have updated this with 32854e0b68a9d03ac3d2e95e40bbe2434c12c7ee.

I did this as a separate commit to this PR as we aren't going to have it as a default attribute and this PR was against the `2.2` branch rather than `develop`.

Thank you for the PR though!

PR #1165, opened 2019-12-06
Fixed product validation error

tim-pixeldeluxe:
Fixed a product validation error when a variant didn’t have a sku: Return value of craft\commerce\elements\Variant::getSku() must be of the type string, null returned


lukeholder:
fixed in 1f324eb

PR #1194, opened 2019-12-31
Change E-Mail formfield to type="email"

kaspar-allenbach:
Is there a reason this is `type="text"` and not `type="email"`?

With `type="email"` you can apply [html5 frontend validation](https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation) much nicer


lukeholder:
Thanks!

PR #1256, opened 2020-01-31
fixed error if the tax is negative

MaximilianSchwarz:
If the amount is negative the tax adjuster calculate negativ taxes.

For example:
- Put one product in the cart. e.g (Product A 400$)
- Create a new coupon code for the product with 500$ off and apply the coupon in the cart page.
- Now craft commerce calculates negative tax. 400$ - 500$ = -100$
- Now craft calculates the tax with -100$

In this pull request the tax adjuster is fixed. I have created a new function to check if the tax is negative. If the tax is negativ the amount of the tax adjuster set to 0.

### DIFF
--- a/src/adjusters/Tax.php
+++ b/src/adjusters/Tax.php
@@ -118,6 +118,17 @@ class Tax extends Component implements AdjusterInterface
         return $adjustments;
     }
 
+    private function _negativeTax($amount)
+    {
+        if($amount > 0) {
+            $status = false;
+        } else {
+            $status = true;
+        }
+
+        return $status;
+    }
+
 
     /**
      * @param TaxRate $taxRate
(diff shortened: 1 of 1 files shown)


lukeholder:
You want to make any tax amount: `max($amount,0);`? Thinking that might be a better solution, but will need to test.

MaximilianSchwarz:
@lukeholder Thanks for the reply.

Right. If the cart amount smaller than 0 we always set the tax amount to 0. It make no sense to calculate tax if the cart amount is 0 or negative.

I think `max($amount, 0)` is the better solution, so you don't need a separate function to check it.

lukeholder:
This has been fixed with Commerce 3.1.0

Discounts can no longer make the order negative.

PR #1258, opened 2020-01-31
Many more german translations

MisterMike:
I have made a few german translations and changed a few to more fitting (IMO) strings.


lukeholder:
Thankyou so much but can you please contribute to Commerce translations with Crowdin: https://crowdin.com/project/craft-commerce

This is the better way to contribute these. Thanks!

PR #1271, opened 2020-02-07
Fixes actionButton on subscription plans in CP

tnerbovik:
This fixes the miss-spelling of «actionButton» block, current spelling is «activeButton», which causes the index page on Subscription plans to display no button to add new plans.


lukeholder:
Thanks!

PR #1274, opened 2020-02-08
Fixes JavaScript error #1273

### LINKED ISSUE #1273: Variant JavaScript error in Craft Commerce 3
### Description

Products with no variants, which I believe is now the default, seems to cause a JavaScript error in Craft 3.4 and Craft Commerce 3 in the control panel.

Screenshot 2020-02-07 16 34 43

I’m thinking this line might be causing the problem: https://github.com/craftcms/commerce/blob/c21f82d06ead3161de2ef9f3b00c29b14d1b89ac/src/templates/products/_edit.html#L246

### Steps to reproduce

1. Add a new User select field to a Product
2. Add a new Product, of a Product Type that doesn’t have Variants. You’ll be able to use the User select field.
3. Edit that product. You won’t be able to use the User select field, as the JS won’t work due to this error.

### Additional info

- Craft version: 3.4.5
- PHP version: 7.3
- Database driver & version: pgsql
- Plugins & versions: Commerce 3.0.6

This is a default Craft CMS (3.3 → 3.4) and Craft Commerce (2.x → 3.x) install, using the default products.

kennethormandy:
This fixes #1273 from the template side of things. It seems like `productType.hasVariants` can return `''` in this situation, so maybe there is a better place to fix it rather than this check in the template.

### DIFF
--- a/src/templates/products/_edit.html
+++ b/src/templates/products/_edit.html
@@ -243,7 +243,7 @@
 
       new Craft.Commerce.ProductEdit({
         id: {{ product.id|raw }},
-        hasVariants: {{ productType.hasVariants|raw }},
+        hasVariants: {{ productType.hasVariants is defined and productType.hasVariants ? productType.hasVariants|raw : 'false' }},
         purchasables: {{ product.getVariants()|map(v => {id: v.id, title: v.title, sku: v.sku})|json_encode|raw }}
       });
   {% endjs %}


lukeholder:
Thanks for the PR, this has been fixed in https://github.com/craftcms/commerce/commit/d46a5d59714a7629b2da20a282059c69572c7588

PR #1290, opened 2020-02-20
Product + Subscription Hooks

AugustMiller:
This is rather self-serving, but: I'm working on something that could benefit from being able to inject a bit of information into the Subscription edit page, and thought I'd issue a PR with the patch in case y'all were sympathetic! ❤️

All this adds is three hooks:

1. Product "Content" area (unfortunately, appearing below any active tab content?)
2. Subscription "Content" area
3. Subscription "Details" area

I didn't see a section in the docs to mention these, so I haven't made any corresponding changes there.

✌️

### DIFF
--- a/src/templates/products/_edit.html
+++ b/src/templates/products/_edit.html
@@ -153,6 +153,8 @@
             </div>
         </div>
     {% endif %}
+
+    {% hook "cp.commerce.product.edit.content" %}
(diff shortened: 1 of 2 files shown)


lukeholder:
Added 8548ff3d051c11e85066f941720f496fb44c6305

PR #1291, opened 2020-02-20
Fixed Product editor unknown error

jmauzyk:
Was getting an unknown error when trying to access the Product editor by double clicking a product in a relational field. Looks like the `generalVariantFields` macro was getting passed a variant but not a product as well.


nfourtythree:
Hey @jmauzyk 

Thank you for this, I have merged it in and it will be included in the next release.

Thanks

PR #1315, opened 2020-03-04
Add ability to query variants from GraphQL

nfourtythree:


mattstein:
Added https://github.com/craftcms/commerce-docs/tree/feature/gql-products-variants to include GQL query examples.

nfourtythree:
@mattstein did you come across any issues when using this branch? Or did everything seem pretty safe to merge?

mattstein:
@nfourtythree I tested each example I added and everything worked great. I figure the source docblocks will also need updating to add GQL examples, which will be the bigger task assuming they’re each tested. Can work on that if you’d like.

PR #1316, opened 2020-03-04
Allow template path changes from EVENT

importantcoding:
The previous code disregarded the $event->template path and only referenced $templatePath rendering any changes to the template path in EVENT_BEFORE_RENDER_PDF useless.


lukeholder:
Thanks!

PR #1331, opened 2020-03-20
Adds “Twig code” instructions to email recipient

kennethormandy:
I’m looking to dynamically fill the Recipient field email address using Twig. From a quick test, that seems to already be supported.

This Pull Request adds the same instructions as on the CC and BCC fields, so it’s clear it behaves the same.

These lines aren’t translated, correct?

Thanks!


kennethormandy:
After another look—I think I was wrong about this.

Is this intentional/necessary, or would it be possible to add the same Twig code feature to the Recipient field?

PR #1370, opened 2020-04-09
fix event class in example

billythekid:


keyurshah:
We just encountered this. Awesome!

nfourtythree:
Thanks for this!

PR #1418, opened 2020-04-30
Use currency formatter defaults

boboldehampsink:
So i.e. `priceAsCurrency` returns the same as `price | currency`


lukeholder:
This should be solved in Commerce 3.1.11

- Added new `*AsCurrency` attributes to all currency attributes on orders, line items, products, variants, adjustments and transactions.

both the `*AsCurrency` attributes and the twig filter use the same helper method in `\craft\commerce\helpers\Currency::formatAsCurrency`

PR #1443, opened 2020-05-11
Make product sources' labels translatable

sosov1:
This PR adds support for translations in the side view of the product's element index table.
I'm not sure if the `commerce` translation category is the correct one, or if `site` would be better suited.

### Before

image

### After

image

### DIFF
--- a/src/elements/Product.php
+++ b/src/elements/Product.php
@@ -969,7 +969,7 @@ class Product extends Element
 
             $sources[$key] = [
                 'key' => $key,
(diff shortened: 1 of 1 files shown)


nfourtythree:
Hi @sosov1 

Thank you for this, as you pointed out it needed to be `site` as it is information entered by a user.

Have fixed this in the following commit: 669a5e8afe9e38a2bcec8094cbecaf5646cbe6cd

Thanks!

PR #1450, opened 2020-05-14
Fixed typo in getTaxCategoryByHandle docs

jorenvandeweyer:
`@param int $taxCategoryHandle`
int should be string


lukeholder:
Thanks!

PR #1461, opened 2020-05-21
Fix namespace of NotFoundHttpException on product

billmn:
Before this fix it returns:
```
Class 'Symfony\Component\HttpKernel\Exception\NotFoundHttpException' not found
```


lukeholder:
Oops. Thanks!

PR #1462, opened 2020-05-22
GraphQL: added “isAvailable” field in Variant schema

billmn:
Could be very useful use the existent `isAvailable` method instead of make all checks manually (status, stock, product available for purchase, ...).

https://github.com/craftcms/commerce/blob/develop/src/elements/Variant.php#L922


nfourtythree:
Thanks!

Issue #611, opened 2018-12-20
Add a cancel() method to Gateway Interface / Gateways


jellingsen:
### Description

When creating a Payment Gateway you can capture and refund, and you can authorize. But you can't unauthorize or cancel a payment.

For the payment provider that I currently is creating a Gateway for, un-captured payments will stay for several weeks before they are auto-cancelled.

lukeholder:
@andris thoughts?

jaydensmith:
No further update on this?

lukeholder:
- [ ] Should be able to cancel uncaptured payment if gateway supports it.

sjelfull:
@lukeholder Is this something you are working on, or should I take a stab at at PR?

sjelfull:
Just following up: Our use case is that all our refunds is cancellations of the payment intent instead of the already supported refund.

This is because payment have to happen on or after delivery, while most support cases happens before, usually because they did not get a delivery.

lukeholder:
@sjelfull you are welcome to make a pull request.

Since you will need to add a `cancel()` and `supportsCancel()`method to the gateway interface this will be a breaking change for gateways that do not inherit from base gateway (if any) so would leave this for a point release.

lukeholder:
Also would need to think about adding something to a transaction to know whether it can still be cancelled (time limits for cancellation by gateway usually) so a `canCancel()` method to transaction model, and `canCancelTransaction()` method to the transaction service.

lukeholder:
@sjelfull Just thinking about the implementation of this.

Instead of changing the current `GatewayInterface`, we would want to create an additional `CancellableGatewayInterface` with the following 2 methods:

`cancel(Transaction $transaction) : RequestResponseInterface`
`canCancel(Transaction $transaction) : bool`

Then in Commerce 4 we will split out the other optional features of a gateway into their own interfaces. (Like partial refund).

You would still add `canCancel()` to the transaction model which calls the transaction service's `canCancelTransaction()`. That latter method would see if the gateway implements the CancellableGateway interface (which saves us having to call a `supportsCancel()` method, and also calls `canCancel(Transaction)` to know it cancellation is even possible.

Let me know if you have any questions.

PR #1480, opened 2020-05-29
Added void action for transactions

jmauzyk:
Taking a crack at creating a way to void transactions as requested in #611, using Omnipay `void()` method.

When transactions have not yet been captured, the time period in which they can be voided varies by card issuer, so I've it set up to provide the Void action for any transactions that have not yet captured.

Once a transaction has been captured, the time period that a transaction can be voided in depends on the gateway and when they batch them, but generally is about 24 hours. I added a `voidWindow` property to the base Commerce Omnipay gateway (craftcms/commerce-omnipay#13), which defaults to 24 hours and can be overridden by each gateway plugin.

I'm not very familiar with PostgreSQL, so for the migration I patterned it after a similar column change in a previous migration.

I tested using [jmauzyk/commerce-cardconnect](https://github.com/jmauzyk/commerce-cardconnect) and [craftcms/commerce-paypal](https://github.com/craftcms/commerce-paypal) and was able to successfully void transactions in both.

Thanks for your consideration!


lukeholder:
Thanks for the PR. Will review.

I think this solves https://github.com/craftcms/commerce/issues/611

lukeholder:
Thanks for your start on this @jmauzyk  - needed to make a few changes and working on it on this branch: https://github.com/craftcms/commerce/pull/1579

Issue #44, opened 2017-03-14
Inventory management


brandonkelly:
- Manage stock on a per-location basis 
- Allow backorders
- Stock screen tracks adding and deduction of stock.
- Allow Bulk updating of stock.
- Track number of purchasable sales over a period of time.
- Allow reservation of stock while items are in the cart. ?
- Change 'unlimited stock' to 'track stock' ?
- Update stock when manually editing the order on order edit page.
- Update stock option when deleting an order.

MichaelWheldon:
Just checking on the status of `Out of stock threshold #50`. This would be a great feature.

meinhard-meintjes:
Hello everyone, I'm also currently trying to figure out how to handle the lack of this feature. Any updates on the status? If it's still a long time away from being ready, is the easiest/best solution writing a plugin or running a separate instance/DB (in this case 2 instances would be enough) in your opinion? Thanks!

jacobrussell:
@lukeholder @brandonkelly Can you provide any update on the status of this issue?  Is it in your pipeline or still future development?

lukeholder:
@jacobrussell Was and is, but we can't give a timeline on it sorry. Should be in Commerce 4.

denisyilmaz:
Is there a plan to have the backorder-feature arriving in Commerce 3?

lukeholder:
@denisyilmaz no plan in Commerce 3 sorry.

denisyilmaz:
is there any update on a timeline when commerce 4 will be available? some of these features are urgently requested by some projects and it would be very nice to give the project-owners a idea of when these features can be revisited / or if a solution must be found in any other way.

jacobrussell:
@lukeholder Can you provide any update here?  Is this expected at some point this year?  Solid inventory management is a core e-commerce feature that's hard to backfill with custom solutions well.  Would be very helpful to know if we can still expect this with Commerce 4.0, and if that's coming soon.

lukeholder:
@denisyilmaz @jacobrussell we are beginning work on multi-store support, and stock/inventory system to support each store is a core feature. Sorry for the delay but we are working on it. Sorry I can't give a timeline.

CMeldgaard:
@lukeholder Is creating a custom stock-field the only fix to make backorder possible in Commerce 3 ?

mediabeastnz:
👀

PR #1516, opened 2020-06-17
Added get variant stock event

samuelbirch:
This will allow the ability for plugins to say how much stock is available for a variant. We can achieve negative stock/back ordering/etc using this.

I've added the EVENT_AFTER_REMOVE_LINE_ITEM to the_saveLineItems function in the Order element.

### DIFF
--- a/src/elements/Order.php
+++ b/src/elements/Order.php
@@ -2792,7 +2792,15 @@ class Order extends Element
         // Delete any line items that no longer will be saved on this order.
         foreach ($previousLineItems as $previousLineItem) {
             if (!in_array($previousLineItem->id, $currentLineItemIds, false)) {
-                $previousLineItem->delete();
+				$lineItem = Plugin::getInstance()->getLineItems()->getLineItemById($previousLineItem->id);
+				$previousLineItem->delete();
+				
+				// Raising the 'afterRemoveLineItemToOrder' event
+				if ($this->hasEventHandlers(self::EVENT_AFTER_REMOVE_LINE_ITEM)) {
+					$this->trigger(self::EVENT_AFTER_REMOVE_LINE_ITEM, new LineItemEvent([
+						'lineItem' => $lineItem,
+					]));
+				}
             }
         }
 
--- a/src/elements/Variant.php
+++ b/src/elements/Variant.php
@@ -166,7 +166,26 @@ class Variant extends Purchasable
      * );
      * ```
      */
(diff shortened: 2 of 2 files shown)


lukeholder:
Thanks Samuel, please give me some time to look through all this.

samuelbirch:
Hey @lukeholder, any thoughts on the get variant stock event?

lukeholder:
Sorry I am away from keyboard at the moment and was actually going to write a reply about the event when back. Give me a couple hours. From mobile.

lukeholder:
Didn't realise my commit close this pull request￼.

samuelbirch:
@lukeholder no problem :)

lukeholder:
I have added 2 events to help with knowing when a line item was added and removed from the order in a clearer way with commit 7baba4a

We don't want `getStock()` to be a getter with an event at the moment. When we add an inventory management system in the future (https://github.com/craftcms/commerce/issues/44), we will give plugins the opportunity to extend the system then.

Putting an event in the getter doesn't really make sense in the case of our basic stock counter being deducted when items are purchased. If you are hitting an external API to get the stock, probably best to set the stock to unlimited and add a custom field that manages your own stock (You can even add your own validator to the line item to check that stock).

If your plugin/module wants to add stock to a product, best to do something like this:

```php
// Add 9999 stock to a variant.
Craft::$app->getDb()->createCommand()->update(
Table::VARIANTS,
['stock' => new Expression('stock + :qty', [':qty' => 9999])],
['id' => $variant->id]
)->execute();
```

If you want to explain your use-case more, happy to talk more about it here.

Thanks.

samuelbirch:
Thanks @lukeholder. I'm writing a inventory management plugin at the moment which does manage the stock separately. I'm needing a way to check that stock when for a line item is added/edited but also for admin orders in the variant element picker.

if i set everything to unlimited then in admin orders it would all be available but i still need to be able to say which is and which isn't available. Is there a solution for this?

Anubarak:
Thanks for this pull request, I made a custom stock system as well and it's per "store" (think about it like in a restaurant)
And this logic helped me to get rid of the Events on record basis where I had to store the lineItem and the correct order during other events.

One Question: Any plans to trigger those functions, when an Order is deleted?
Currently I cannot increase my stock again unless I use the afterDelete Event of the order but it would be more convenient if deleting an order would trigger the event as well in it's custom "beforeDelete" function

PR #1526, opened 2020-06-19
Add setting to optionally process payment on free orders

johnnynotsolucky:
We have a payment processor which maintains customer balances and customers can have a positive balance. When the customer makes a purchase, the processor uses this balance as part of the payment. In Craft, we have an adjuster which fetches the existing balance and applies it like a discount on the order.

In some (rare) cases, customers can have enough credit that their balance covers the cost of the order. In these instances, Commerce applies the adjustment, but as the order doesn't have an outstanding balance, it completes the order without forwarding to the payment processor. The customers balance in the processor remains the same and they can keep placing orders below that amount.

This PR adds a configuration option to allow users to prevent "free" orders from being completed, and instead forwarded to the payment gateway. It still defaults to complete.

I'm not sure if this is the best way to do this, or whether there is another way around it without code changes.


sjcallender:
@lukeholder Would you be able to take a look at this and tell us if this looks like something you'd be willing to add soonish? We'd prefer a first-party solution, but can roll a temporary hack sooner if we know you are not prioritizing this.

src/models/Settings.php
```diff
@@ -93,6 +96,16 @@ class Settings extends Model
...
 
+    /**
+     * @var string
+     */
+    public $freeOrderPaymentStrategy = 'complete';
+
+    /**
+     * @var bool
+     */
+    public $mergeLastCartOnLogin = true;
```
lukeholder, line 107: Merged thsi PR, but `mergeLastCartOnLogin` is an old config setting from commerce 2 which I removed.

Thanks.

lukeholder:
@sjcallender Will be in the next Commerce release. Thanks.

sjcallender:
Thank you, @lukeholder.

PR #1537, opened 2020-06-25
Added a transaction cache on orders

lukeholder:
Before transaction (sum) refactor: 37.8 seconds

image

After transaction refactor (removal of sum transactions query) : 15.8s

image

- Move the total paid/refunded/authorized to order model and out of payments service
- Transactions cached on order model
- Transactions cleared when updateOrderPaid is called on the order. This would/should be done by anyone creating transactions on the order so it gives us the oppotunity to clear the cache
- Fixed the order total in a customer’s orders with admin table

### DIFF
--- a/src/services/Discounts.php
+++ b/src/services/Discounts.php
@@ -541,12 +541,14 @@ class Discounts extends Component
             return false;
         }
 
-        $orderDiscountConditionParams = [
-            'order' => $order->toArray([], ['lineItems.snapshot', 'shippingAddress', 'billingAddress'])
-        ];
+        if ($discount->orderConditionFormula) {
+            $orderDiscountConditionParams = [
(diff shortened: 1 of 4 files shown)


src/services/Discounts.php
```diff
@@ -541,12 +541,14 @@ public function matchOrder(Order $order, Discount $discount): bool
             return false;
         }
 
-        $orderDiscountConditionParams = [
-            'order' => $order->toArray([], ['lineItems.snapshot', 'shippingAddress', 'billingAddress'])
-        ];
+        if ($discount->orderConditionFormula) {
```
lukeholder, line 544: @nfourtythree this was a massive performance improvement. Serialising the order every time we matched a discount was very expensive.

PR #1546, opened 2020-07-01
Allow multi keyword variant searching

samuelbirch:
This allows multi keyword searching of variants when adding a lineitem in the edit orders page

eg: a variant with the description: Burgtec Penthouse MK5 Flat Pedals Steel Axle Iron Bro Orange

can now be found using the search phase: burgtec mk5 pedal orange


lukeholder:
Thanks!

Issue #1560, opened 2020-07-09
Move the “Categories Relationship Type” discount setting under an “Advanced” toggle


brandonkelly:
The “Categories Relationship Type” setting is pretty [confusing](https://discordapp.com/channels/456442477667418113/672206118122160149/730671617826357259) to shop owners, and also pretty niche. I imagine 99% of use cases would be happy with the behavior you get when “Both” is selected.

So I think it should be moved under an “Advanced” toggle, which is hidden by default, similar to what we do in Matrix and Redactor field settings, among others. That gives shop owners permission to ignore it unless they have a good reason to care.

While you’re at it, the instruction (and possibly option label) copy should be updated for clarity. At the very least the reference to “Products” should be replaced with “purchasables”, and “categories” should be lowercased.

lukeholder:
Fixed in https://github.com/craftcms/commerce/pull/1565

PR #1565, opened 2020-07-13
Initial work on making it clearer how category relationships work with promotions.

lukeholder:
Welcome @mattstein to make it more clear in the field instructions - will then extend this to sales.

### DIFF
--- a/src/controllers/DiscountsController.php
+++ b/src/controllers/DiscountsController.php
@@ -377,10 +377,15 @@ class DiscountsController extends BaseCpController
 
         $variables['categories'] = $categories;
 
-        $variables['categoryRelationshipType'] = [
-            DiscountRecord::CATEGORY_RELATIONSHIP_TYPE_SOURCE => Plugin::t('Source'),
-            DiscountRecord::CATEGORY_RELATIONSHIP_TYPE_TARGET => Plugin::t('Target'),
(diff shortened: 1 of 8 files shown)


lukeholder:
Fixed #1560

lukeholder:
Also fixed the migration for people upgrading to Commerce 3. They should stay on sourceElement relation type.

lukeholder:
@mattstein merged it for this release, but welcome your input still on the option labels.

mattstein:
Sorry I didn’t respond sooner, @lukeholder!

I did a quick mockup thinking through this, and while it’s probably annoying scope creep I think it’d be most useful to illustrate how the relationship match works:

category-promotion-mockup

This is a challenging thing to conceptualize, so using a radio field puts the options side by side for the user to compare and consider. Simply illustrating the options could potentially make it easier to understand, save some language, and potentially avoid a trip to the (more technical) docs.

I tried to also focus the language from the perspective of the categories, since we’re choosing them here, which could be problematic since I think the wording of the current options is focused on the purchasables. (Note how the illustrations are the inverse of what they describe.)

Happy to chat further or refine, or just run and hide if this is TMI! :)

mattstein:
Also I’m aware that’s pretty hideous so please consider it an illustration and not design work.

PR #1579, opened 2020-07-20
Allow gateway transactions to voided

lukeholder:
Start of voiding transactions with payment gateways.

- Added a `Gateway::supportsVoid()` and `Gateway::void()` to Gateway interface and base abstract class.
- Implemented support to the Dummy gateway

- [ ] Figure out how to add a void window setting to gateways that support it.
- [ ] Test voiding with a real gateway that supports it.


jmauzyk:
Wanted to restart the conversation about adding this feature. Would be really great to be able to void transactions and have a record of it in Commerce. Is this still a possibility?

PR #1590, opened 2020-07-26
Add EVENT_BEFORE_DELETE_ADDRESS event

johnnynotsolucky:
### Description

This PR adds the `EVENT_BEFORE_DELETE_ADDRESS` to the Addresses service and allows for instances where you'd need to do some work based on the customer or some other relation to the address which would be deleted when the address record is deleted.

For example, with this change if we're caching a customers addresses, we can cache that data indefinitely and only invalidate it when an address linked to that customer is deleted (or updated).

### Related issues

This adds onto https://github.com/craftcms/commerce/pull/810 from last year.

### DIFF
--- a/src/services/Addresses.php
+++ b/src/services/Addresses.php
@@ -86,6 +86,29 @@ class Addresses extends Component
     const EVENT_AFTER_SAVE_ADDRESS = 'afterSaveAddress';
 
     /**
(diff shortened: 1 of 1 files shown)


lukeholder:
Thanks for that. Sorry for the delay!

PR #1689, opened 2020-08-31
Fixing twig example code (changed endif to endfor)

GaryReckard:
### Description
Noticed a small typo in the example twig code. Replaced the errant `endif` with `endfor`.


nfourtythree:
Thanks @GaryReckard

PR #1705, opened 2020-09-03
Check if coupon availaible with discount condition formula

fhamon:
#### The problem
Craft's commerce does not return an error when a user wants to add an existing coupon code to the cart via the action `commerce/cart/update-cart` and the order condition formula did not pass.

#### The solution
I added this simple check based on your code style and existing code to validate and send an error when the coupon is not valid because of the condition formula.

Please let me know if this PR is problematic, I am ready to change the code or adapt to your guidelines.

Thanks for Craft and thanks for the commerce plugin. Much love from Montreal ❤

### DIFF
--- a/src/services/Discounts.php
+++ b/src/services/Discounts.php
@@ -365,6 +365,11 @@ class Discounts extends Component
             return false;
         }
 
+        if (!$this->_isDiscountConditionFormulaValid($order, $discount)) {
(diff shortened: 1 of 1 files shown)


lukeholder:
Thanks for that!

lukeholder:
Changelog entry: 3af212ed836f7724043071846212e745aa353f0a

PR #1740, opened 2020-09-17
Document Settings

mattstein:
Preparing the settings model to auto-generate documentation by alphabetizing properties and making sure each one is explained:

- Alphabetizes public properties.
- Adds or updates documentation for each property.
- Removes unused `$paymentMethodSettings`.

**Not ready to be merged!** Doc generator needs revision + testing first.


src/models/Settings.php
```diff
@@ -40,191 +45,317 @@ class Settings extends Model
...
+     * @group Orders
      */
-    public $validateCartCustomFieldsOnSubmission = false;
+    public $useBillingAddressForTax = false;
 
     /**
-     * @var string|null the uri to redirect to after using the load cart url
-     * @since 3.1
+     * @var bool Whether business tax ID should be validated separately from VAT ID.
+     *
```
lukeholder, line 327: @mattstein this setting is just deciding to add input validation to the address attribute `businessTaxId`.

If you have a vat-based tax rate, it will look at the addresses `businessTaxId` to figure out if the ID is a valid VAT ID from an API and not apply the VAT tax as an adjustment. But this isn't going to stop someone from putting anything they want into the `businessTaxId` field on the address. 

Some people wanted an address validation error when the person put in a bad vat ID. let me know if that makes sense.

lukeholder:
Everything else looks good.

src/models/Settings.php
mattstein: Thanks @lukeholder. If I’m understanding correctly:

- `false` means Commerce won’t validate `businessTaxId`, which is ideal if I’m not dealing with VAT, not collecting a `businessTaxId`, or handling VAT with my own custom logic (?).
- `true` means Commerce will make sure `businessTaxId` is a valid VAT ID or throw a validation error.

>  But this isn't going to stop someone from putting anything they want into the businessTaxId field on the address.

You mean when it’s `false`? Because it looks like if this was `true` and I entered `handsomeluke` as my `businessTaxId`, the [resulting validation error](https://github.com/craftcms/commerce/blob/da7b828f13beb74a29c015c707ff71d6d91cda02/src/models/Address.php#L392-L392) would require me to provide a valid ID.

> if the ID is a valid VAT ID from an API and not apply the VAT tax as an adjustment

Pardon my ignorance, but if the store collects VAT and I (the customer) provide a VAT ID, any VAT charges on my order will be removed? Presumably because my ID will be used to later determine what, if anything, I owed for tax on that order?

mattstein:
[Merged Commerce #1740](https://github.com/craftcms/commerce/pull/1740) with Luke’s blessing, so this is ready to go now.

Issue #1379, opened 2020-04-13
Order emails should no longer catch errors


AugustMiller:
Now that order emails are sent via the Queue, failure can be tracked in other ways (aside from logging the exception messages).

Debugging would be _much_ simpler if you could visualize what emails failed to send, fix the error, and then re-try, rather than failing quietly… but still completing in the queue!

Basically, this chunk of logic to swallow template processing errors feels a bit out of place, in the new model, and could probably just be allowed to blow up and report exceptions to the Queue: https://github.com/craftcms/commerce/blob/develop/src/services/Emails.php#L430-L849

In other words—now that we have a means to record and recover from failures, we should probably do that, rather than hiding rendering/deliverability issues.

lukeholder:
Great idea. Will do.

gojira32:
Any progress on this. We just got caught out with this.

> Debugging would be much simpler if you could visualize what emails failed to send, fix the error, and then re-try, rather than failing quietly… but still completing in the queue!

This would have been super handy...

lukeholder:
Thanks for your PR @AugustMiller - you were right though, it was breaking change to the `sendEmail()` expecting to only return a bool.

I have gone ahead made an alternative fix in 383f53b0db4ede34d4ccaea6154483cdaf6b3ac2

Will be in the next release. Thanks!

PR #1779, opened 2020-10-11
Improve error reporting when sending Order Emails

AugustMiller:
This addresses #1379. I've felt for a while now that this is a necessary change for developers and administrators to know where issues are occurring with their customer emails.

Instead of swallowing errors (and only reporting them to the log), this lets them bubble out of `sendEmail`, either halting the parent `SendEmail` job, or the `OrderController::actionSendEmail` method.

As a result, emails that fail to send are either:

1. Immediately reported to the user, when using the "Send Email" function
2. Kept in the queue for review

The reported errors are more general (but still reflect whatever aspect failed)—perhaps the documentation can be amended, to suggest looking in the logs for more information?

Screen Shot 2020-10-11 at 12 00 58 AM

I've used the existing `craft\commerce\errors\EmailException` class, but happy to create more specific errors for template/path, address, rendering, and/or PDF issues.

This was a one-shot attempt—let me know if it needs a style review, or if you've got something in the pipeline already. 😉

_It might be possible to let the individual exceptions bubble up (i.e. no `try/catch`), but I figured that a way of consolidating errors in a way that was meaningful to the admin/developer was better than pinpoint accuracy in a job or flash message…_

### DIFF
--- a/src/controllers/OrdersController.php
+++ b/src/controllers/OrdersController.php
@@ -578,15 +578,9 @@ class OrdersController extends Controller
 
         $success = true;
         try {
-            if (!Plugin::getInstance()->getEmails()->sendEmail($email, $order)) {
-                $success = false;
-            }
+            Plugin::getInstance()->getEmails()->sendEmail($email, $order);
         } catch (\Exception $exception) {
-            $success = false;
-        }
-
-        if (!$success) {
-            return $this->asErrorJson(Craft::t('commerce', 'Could not send email'));
+            return $this->asErrorJson($exception->getMessage());
         }
 
         return $this->asJson(['success' => true]);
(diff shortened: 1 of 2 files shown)


AugustMiller:
Also worth noting (mentioned in craftcms/docs#87): When a plugin surpasses an email, an exception is _not_ thrown. It's up to the calling method to test for a `false`y result.

In thinking more about this, **it might be considered a breaking change**—if someone is calling `sendEmail()` directly in their application code, this will start throwing exceptions instead of just failing silently. Yikes?

MattWilcox:
+1 for this; the amount of time we wasted faffing with Email thinking we had an issue with SendGrid integration or something, when it was just Commerce failing to give feedback on time-critical errors involving customer data...

lukeholder:
This has been fixed in https://github.com/craftcms/commerce/commit/383f53b0db4ede34d4ccaea6154483cdaf6b3ac2

Thanks for the PR!

Issue #581, opened 2018-11-23
Reset all commerce data command (CLI)


goellner:
After testing and development on a local machine, it would be nice to clear the orders and customers, to have a clean database on production.

Is there any other way to achieve that in the meantime?

lukeholder:
Deletion of all Orders would be done in PHP like this:

```
$query = Orders::find();
$elementsService = Craft::$app->getElements();
foreach ($query->limit(null)->all() as $element) {
     $elementsService->deleteElement($element);
}
```

Deletion of all customers:

```
\craft\commerce\records\Customer::deleteAll();
```

Will keep the ticket open as an enhancement.

lukeholder:
#5

lukeholder:
See #1043

lukeholder:
See the https://craftcms.com/docs/commerce/4.x/console-commands.html#reset-data

PR #1781, opened 2020-10-12
Reset Data Console Command

nfourtythree:
### Description
This adds a console command for resetting Commerce data. The command takes the following actions:

- Deletes all orders/carts
- Deletes all subscriptions
- Deletes all payment sources
- Deletes all guest customers
- Deletes all addresses (for both guest and user customers)
- Resets all discount usage counters (total uses, customer usage and email usage)

This is useful for people developing to be able to clear any data before migrating the database.

A prompt is given before the reset action takes place.

### Related issues
#581

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,12 @@
 ## Unreleased
 
 ### Added
+- Added the `commerce/reset-data` console command. ([#581](https://github.com/craftcms/commerce/issues/581))
 - Added a “Variants” column to the Products index page. ([#1765](https://github.com/craftcms/commerce/issues/1765))
-- Added `craft\commerce\services\PaymentSources::getAllPaymentSourcesByGatewayId()`.
-- Added `craft\commerce\helpers\Purchasable`.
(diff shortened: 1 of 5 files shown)


lukeholder:
Awesome thanks

KatieMFritz:
I notice there are still rows in my `craft_commerce_customers_addresses` table after I run the command. Where are those coming from?

nfourtythree:
Hi @KatieMFritz 

The `commerce_customers_addresses` table has cascade deletes for when an address or a customer is deleted. So these records should be removed when the parent data is removed.

Do the customer IDs in the table actually still have records in the `commerce_customers` table?

KatieMFritz:
Sorry for the delayed response!

Yes, the `commerce/reset-data` command leaves customers. It does delete orders.

KatieMFritz:
Oh, is it maybe because they're not guest customers?

nfourtythree:
Hi @KatieMFritz 

All users have a customer record. These customer records do not get deleted when using the `reset-data` command.

Thanks!

PR #1797, opened 2020-10-22
Follow suit of Craft's change to label of Share button

Saboteur777:
Fix https://github.com/craftcms/cms/issues/6963 here, too.


nfourtythree:
Hi @Saboteur777 

Thanks for this!

Issue #1809, opened 2020-10-24
404 error when creating shipping zone


fantasticmachine:
404 error when trying to create a shipping zone.

Create a shipping method.
Create a shipping rule
Under Conditions, try to create a new Shipping Zone.
The form that appears in the modal tries to submit to  `/index.php?p=admin/actions/commerce/shippingZones/save and causes a 404 error`

I believe the URL should be `/index.php?p=admin/actions/commerce/shipping-zones/save`

**Additional info**
- Craft version: 3.5.14
- Commerce version: 3.2.8.1
- PHP version: 7.3

fantasticmachine:
I submitted a pull request to fix this issue but it appears the build is failing.

lukeholder:
Thanks, I merged the fix!

PR #1810, opened 2020-10-24
Fix incorrect shipping zone submission url

fantasticmachine:
Fixes an incorrect URL for the creation of shipping zones

URL for shipping zone creation form was wrong.

https://github.com/craftcms/commerce/issues/1809


lukeholder:
Thanks! Fixed #1809

PR #1814, opened 2020-10-27
Ignore "Access Craft Commerce" permission on some ProductsController actions.

### LINKED ISSUE #1791: 'Manage Products' permission also requires 'Access Craft Commerce' permissions?
**Description**
We have a front end product form. I've enabled the "Mangage Products" permission for the relevant product type. However when trying to add a product the front end form reports ''does not have permission". The only was I can get it to work, is to enabled the "Access Craft Commerce" permission in General which in turn, also requires CP access. 

Is this right?

**Additional info**
- Craft version: 3.5.11.1
- Commerce version: 3.2.6
- PHP version: 7.3.11

pdaleramirez:
### Description

Fixed #1791

### DIFF
--- a/src/controllers/ProductsController.php
+++ b/src/controllers/ProductsController.php
@@ -40,14 +40,30 @@ use yii\web\ServerErrorHttpException;
  * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
  * @since 2.0
  */
-class ProductsController extends BaseCpController
+class ProductsController extends BaseController
 {
+    /**
+     * @var string[] The action names that bypass the "Access Craft Commerce" permission.
+     */
+    protected $ignorePluginPermission = ['save-product', 'duplicate-product', 'delete-product'];
+
     /**
      * @inheritdoc
      */
     public function init()
     {
         parent::init();
+
+        $actionSegs = Craft::$app->getRequest()->getActionSegments();
+
+        $actionName = $actionSegs[2] ?? null;
+
+        $requireAccessPluginCommerce = ($actionName !== null && in_array($actionName, $this->ignorePluginPermission)) ? false : true;
+
+        if ($requireAccessPluginCommerce === true) {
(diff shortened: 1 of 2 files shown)


src/controllers/ProductsController.php
```diff
@@ -40,14 +40,30 @@
...
+        $actionSegs = Craft::$app->getRequest()->getActionSegments();
+
+        $actionName = $actionSegs[2] ?? null;
+
+        $requireAccessPluginCommerce = ($actionName !== null && in_array($actionName, $this->ignorePluginPermission)) ? false : true;
+
+        if ($requireAccessPluginCommerce === true) {
+            $this->requirePermission('accessPlugin-commerce');
+        }
+
```
lukeholder, line 66: Just realised it might be easier to use the beforeAction method which gives you the action being run directly. Example:

https://github.com/craftcms/cms/blob/develop/src/controllers/UsersController.php#L118-L127

So something like below (untested):

```php
    /**
     * @inheritdoc
     */
    public function beforeAction($action)
    {
        if (!parent::beforeAction($action)) {
            return false;
        }

        if (!in_array($action->id, ['save-product', 'delete-product', 'duplicate-product'])) {
            $this->requirePermission('accessPlugin-commerce');
        }

        if (in_array($action->id, ['save-product', 'delete-product', 'duplicate-product'])) {
            $this->requirePermission('commerce-manageProducts');
        }

        return true;
    }
```

You can then remove the `init()`

lukeholder: Another example in cms repo: https://github.com/craftcms/cms/blob/develop/src/controllers/ProjectConfigController.php#L31-L39

pdaleramirez: I think we don't need to put ```$this->requirePermission('commerce-manageProducts');``` in a condition because we want to have that permission on all the methods

Issue #1799, opened 2020-10-22
Add an event to listen to any webhook request


lukeholder:
**What are you trying to do?**
Use PHP to listen to a webhook that comes into Commerce

**What's your proposed solution?**
Add a before and after event to `\craft\commerce\services\Webhooks::processWebhook`

lukeholder:
Fixed in #1820

PR #1820, opened 2020-10-28
Created events before and after a webhook has been processed.

pdaleramirez:
(#1799)

### DIFF
--- a/src/events/WebhookEvent.php
+++ b/src/events/WebhookEvent.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ */
+
+namespace craft\commerce\events;
+
+use craft\commerce\base\Gateway;
+use yii\base\Event;
+use yii\web\Response;
+
+/**
+ * Class WebhookEvent
+ *
+ * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
+ * @since 3.2
+ */
+class WebhookEvent extends Event
+{
+    /**
+     * @var Gateway
+     */
+    public $gateway;
+
+    /**
+     * @var Response
+     */
+    public $response;
+}
--- a/src/events/WebhookEvent.php
+++ b/src/events/WebhookEvent.php
@@ -15,7 +15,7 @@ use yii\web\Response;
  * Class WebhookEvent
  *
  * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
- * @since 3.2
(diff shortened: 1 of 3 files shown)


src/services/Webhooks.php
```diff
@@ -61,6 +113,13 @@ public function processWebhook(Gateway $gateway): Response
             $mutex->release($transactionLockName);
         }
 
+        // Fire a 'afterProcessWebhook' event
+        if ($this->hasEventHandlers(self::EVENT_AFTER_PROCESS_WEBHOOK)) {
+            $this->trigger(self::EVENT_AFTER_PROCESS_WEBHOOK, new WebhookEvent([
```
nfourtythree, line 118: @pdaleramirez @lukeholder should we include the gateway as well in this event? It could give the developers that extra bit of context.

Thoughts?

src/services/Webhooks.php
```diff
@@ -22,13 +23,64 @@
...
+     *     function(WebhookEvent $event) {
+     *         // @var Gateway $gateway
+     *         $gateway = $event->gateway;
+     *
+     *         // ...
+     *     }
+     * );
+     * ```
+     */
+    const EVENT_BEFORE_PROCESS_WEBHOOK = 'beforeProcessWebhook';
```
nfourtythree, line 46: @pdaleramirez could we add a `@since` tag here please

src/services/Webhooks.php
```diff
@@ -22,13 +23,64 @@
...
+     *     function(WebhookEvent $event) {
+     *         // @var Response $response
+     *         $response = $event->response;
+     *
+     *         // ...
+     *     }
+     * );
+     * ```
+     */
+    const EVENT_AFTER_PROCESS_WEBHOOK = 'afterProcessWebhook';
```
nfourtythree, line 68: @pdaleramirez could we add a `@since` tag here please

src/events/WebhookEvent.php
```diff
@@ -0,0 +1,31 @@
...
+
+use craft\commerce\base\Gateway;
+use yii\base\Event;
+use yii\web\Response;
+
+/**
+ * Class WebhookEvent
+ *
+ * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
+ * @since 3.2
```
nfourtythree, line 18: @pdaleramirez when we add a since tag we normal add it as `@since 3.x` and then just before the next release we find and replace them all to make sure they have the correct version number.

nfourtythree (changes requested):
Have added a couple of my thoughts 👍

lukeholder (approved):

lukeholder:
Thanks @pdaleramirez

PR #1826, opened 2020-10-29
Support variant for set sale price

mostafasoufi:
It could allow setting the sale price during custom initial events, something like this:

```
public function events()
{
    return [
        Purchasable::EVENT_INIT => 'initialEvent',
    ];
}

public function initialEvent(Event $event)
{
    /** @var Variant $variant */
    $variant = $event->sender;

    $variant->setSalePrice(5.99);
    $variant->product->promotable = true;
}
```


lukeholder:
Thanks for this idea. We are discussing internally a path forward with custom price modifications both outside and inside the context of a cart. Keeping this PR open and circle back soon.

mostafasoufi:
Hey, when do think guys you will work on?

Thanks!

stanislavprokopov:
@lukeholder any news on this topic?

lukeholder:
Unfortunately we will be going in a different direction to allow plugins to override the price of the products. Stay tuned to future announcements. Thanks.

PR #1831, opened 2020-10-30
Defer costly n+1 queries in matching sales as much as possible

boboldehampsink:
Try to match on "cheaper" conditions first, before doing expensive n+1 queries

In a certain case of mine the page went from 1400+ queries to ~160 queries


lukeholder:
Thanks for that. We hope to improve the sales system in the future to pre-calculate all sales on a queue job, and remove the lazy sales calculation.

Issue #1836, opened 2020-11-01
Order showing as "Overpaid" - floating point comparison issue


engram-design:
We're getting an odd issue where an order is marked as overpaid, despite the paid value and order value being identical.

This particular order had a partial refund applied. The original order total was $164.90, and a partial refund of $14.95, bringing the new order total to $149.95.

The issue seems to be the floating point number comparison used here https://github.com/craftcms/commerce/blob/develop/src/elements/Order.php#L2149 according to https://www.php.net/manual/en/language.types.float.php#113703 it's not always a great idea to compare floats this way.

I've even gone through some debugging just to check I'm not going crazy:

```
Craft::dump($this->getTotalPaid());
Craft::dump($this->getTotal());
Craft::dump($this->getTotalPaid() > $this->getTotal());

149.95
149.95
true
```

Which is crazy right? Its reporting 149.95 is greater than 149.95. If I hard-code the value of `$this->getTotalPaid()` to 149.95, it actually returns the correct result (false). I recall reading that `var_dump` and the like don't report excessive decimals, which might be why this number may contain hidden values that mess up this comparison.

The solution seems to be adjusting that line to round values (to 2 decimal places, as we're dealing with currency), before comparing.

```
if ($this->getIsPaid() && $this->getTotal() > 0 && round($this->getTotalPaid(), 2) > round($this->getTotal(), 2)) {
```

**Additional info**
- Craft version: 3.3.16.1
- Commerce version: 2.2.23
- PHP version: 7.2

lukeholder:
This has been fixed for the next commerce 2.x release. Keep an eye out.

engram-design:
FYI, we're still seeing this behaviour in 3.3.3. It might have something to do with partial refunds, but there's at least 20 orders where this issue has occurred in the last 6 months, so I'll go through and double check.

For instance, an order's total value is `214.9000`. There are two transactions, one with the amount `214.9000` and `14.9500` - the second being the refund. Changing the refund transaction to `14.9501` fixes the issue, which still showing correct values due to rounding.

codyjames:
We're seeing this as well.

moldedjelly:
We're seeing this issue too. We're using 3.4.14. I did the DB trick @engram-design suggested and it worked as a temp fix.

samhibberd:
We are seeing this in Commerce 4.3.2, impacting maybe 1/1000 orders.

[...]

PR #1837, opened 2020-11-02
Comparing getTotalPaid and getTotal invalid return value.

pdaleramirez:
(#1836 )

### DIFF
--- a/src/elements/Order.php
+++ b/src/elements/Order.php
@@ -1223,7 +1223,7 @@ class Order extends Element
      */
     public function getPaidStatus(): string
     {
-        if ($this->getIsPaid() && $this->getTotal() > 0 && $this->getTotalPaid() > $this->getTotal()) {
+        if ($this->getIsPaid() && $this->getTotal() > 0 && round($this->getTotalPaid(), 2) > round($this->getTotal(), 2)) {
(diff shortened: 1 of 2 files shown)


src/elements/Order.php
```diff
@@ -1223,7 +1223,7 @@ public function getIsPaid(): bool
      */
     public function getPaidStatus(): string
     {
-        if ($this->getIsPaid() && $this->getTotal() > 0 && $this->getTotalPaid() > $this->getTotal()) {
+        if ($this->getIsPaid() && $this->getTotal() > 0 && round($this->getTotalPaid(), 2) > round($this->getTotal(), 2)) {
```
lukeholder, line 1226: We can't hard code 2 decimal places here. It would be better to ensure that the amounts compared are rounded already. When rounding use the Currency helper, which can round to the correct number of decimal places for that currency: e.g: `Currency::round($paid - $refunded, $this->currency);`

lukeholder: Hold off on making another fix as I investigate this one.

lukeholder: For commerce 2, add the rounding inside `getTotalPaid()` and remove it's float return.

lukeholder (changes requested):

Issue #1761, opened 2020-10-01
Throw an event to manipulate DomPDF before rendering a PDF via $dompdf->output()


Anubarak:
**What are you trying to do?**
I'm trying to render images in my order PDF, but DomPDF suddenly changed the way images are loaded. This is explained in https://github.com/dompdf/dompdf/issues/2229
So it would be nice to be able to manipulate your DomPDF object and add
```PHP
$options->setChroot(Craft::getAlias('@assetPath'));
``` 
to be able to render the assets again.

**What's your proposed solution?**
Either create an event to manipulate the path or add a settings variable available via
```PHP
$chrootPath = Plugin::getInstance()->getSettings()->pdfChrootPath;
// ...
// check if path is not empty and such...
$options->setChroot($chrootPath);
```

pdaleramirez:
@Anubarak There is currently breaking change for dompdf 0.8.6 downgrading the previous version would solve the issue for now. Similar issue mentioned here https://github.com/craftcms/commerce/issues/1727

Anubarak:
@pdaleramirez thank you. I know that. I explained it in my first post

pdaleramirez:
Duplicate issue #1727

Anubarak:
Why should that be duplicated?
I just want an event to manipulate the DomPDF Object
It's not only because of that specific issue

pdaleramirez:
@Anubarak dompdf 0.8.6 requires php version 7.1, Craft 3 needs to support 7.0. Craft 4 would raise the php requirement by then we can add the enhancement.

Anubarak:
I know that.. but what does the PHP Version has to do with an additional event to change the DomPDF object in general?

pdaleramirez:
@Anubarak Dompdf states here https://github.com/dompdf/dompdf/releases/tag/v0.8.6 that php requirement is > 7.1.

Anubarak:
I know that... I just don't know why the php requirement for dompdf is an argument against an event in commerce. Maybe @lukeholder can explain why their PHP version prevents you to include an event. Because I can do that without any issues, the version does not matter.

pdaleramirez:
@Anubarak We'll add an event to allow setting of options before it gets rendered.

Anubarak:
Thank you very much 👍

pdaleramirez:
[...]

PR #1844, opened 2020-11-03
Add event to allow modification of dompdf options before render

pdaleramirez:
(#1761)


src/services/Pdfs.php
```diff
@@ -146,6 +147,29 @@ class Pdfs extends Component
...
+     * @since 3.x
+     *
+     * ```php
+     * use craft\commerce\events\PdfRenderOptionsEvent;
+     * use craft\commerce\services\Pdfs;
+     * use yii\base\Event;
+     *
+     * Event::on(
+     *     Pdfs::class,
+     *    Pdfs::EVENT_ADD_RENDER_OPTIONS,
```
lukeholder, line 161: Rename to `EVENT_MODIFY_RENDER_OPTIONS`

lukeholder (changes requested):

src/services/Pdfs.php
pdaleramirez: Done renaming the event name

nfourtythree (approved):
This all looks good to me, as long as the event is not cancellable which it doesn't look like it is 👍

lukeholder:
Thanks!

PR #1845, opened 2020-11-03
Fix postgres migration

### LINKED ISSUE #1843: 3.2.10 is broken on PostgreSQL
**Description**

When upgrading from 3.2.9.1 to 3.2.10 I get the following error:

```
    > execute SQL: ALTER TABLE {{%commerce_variants}} ALTER [[deletedWithProduct]] TYPE bool USING CASE WHEN [[deletedWithProduct]]=1 THEN TRUE WHEN [[deletedWithProduct]]=0 THEN FALSE ELSE NULL END; ...Exception: SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: boolean = integer at character 106
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
The SQL being executed was: ALTER TABLE "commerce_variants" ALTER "deletedWithProduct" TYPE bool USING CASE WHEN "deletedWithProduct"=1 THEN TRUE WHEN "deletedWithProduct"=0 THEN FALSE ELSE NULL END; 
```

**To Reproduce**
How, and what happened?
1. Upgrade from 3.2.9.1 to 3.2.10 on PostgreSQL

**Expected behavior**
A working migration

**Additional info**
- Craft version: latest
- Commerce version: 3.2.10
- PHP version: 7.3
- Database driver & version: PostgreSQL 12.4

**Additional context**
Please test stuff like this before releasing - I keep on submitting fixes after releases

boboldehampsink:
Fixes #1843

### DIFF
--- a/src/migrations/m201102_064231_fix_deletedWithProduct_bool.php
+++ b/src/migrations/m201102_064231_fix_deletedWithProduct_bool.php
@@ -27,7 +27,7 @@ class m201102_064231_fix_deletedWithProduct_bool extends Migration
             // (see https://github.com/yiisoft/yii2/issues/12077)
 
             $this->execute('ALTER TABLE {{%commerce_variants}} ALTER COLUMN [[deletedWithProduct]] DROP DEFAULT;');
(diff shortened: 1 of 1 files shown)


lukeholder:
Released 3.2.10.1 with this fix.

PR #1864, opened 2020-11-10
Unsaved Order edit changes warning

nfourtythree:
### Description
Added extra check on top of the default browser checks to warn a user about unsaved changes when trying to navigate away from the edit order page.


pdaleramirez (approved):
I checkout and tested. It worked.

lukeholder (approved):

PR #1865, opened 2020-11-10
Address validateState doesn't run if stateId/countryId is a numeric string

michaelramuta:
### Description
Change is_int to is_numeric when checking if validateState should be run


lukeholder:
Thanks!

PR #1880, opened 2020-11-17
Added "Cart updated" message to the correct language category

OwenMelbz:
Added "Cart updated" message to the correct language category


lukeholder:
Thanks!

PR #1885, opened 2020-11-19
Fixed notice when a Customer Address was saved

billmn:
Flash Notice is empty when an address was saved successfully.


lukeholder:
Thanks!

Issue #1894, opened 2020-11-25
addressLines property of address does not contain businessName and businessTaxId


piotrpog:
Within `addressLines` object of single address there no `businessName` property, and no `businessTaxId` property.

pdaleramirez:
Fixed for next release

PR #1896, opened 2020-11-27
Added missing businessTaxId key in addressLines.

pdaleramirez:
Fixed (#1894 )


nfourtythree (approved):
Looks go to me. I would just update the changlog message to something along the lines of. 

Fixed missing `businessTaxId` in `addressLines` output

PR #1903, opened 2020-12-01
Feature/partial payments

samuelbirch:
Allow partial payments to be made in the CP. (restricted to the maximum of the order value)


lukeholder:
Merged this into a feature branch. Thanks! Will be making a few adjustments, but should be in the next 3.x release!

Thanks.

Issue #174, opened 2018-03-28
Detailed Product Permissions


andris-sevcenko:
> Created by: Alex Collins ([alex@ethercreative.co.uk](alex@ethercreative.co.uk)) on 2017/04/26 06:16:54 +0000
Votes at time of UserVoice import: 3

------
Give product type user permissions the same granularity as a Channel in CMS

andris-sevcenko:
> Created by: Maarten ([wiejeben@gmail.com](wiejeben@gmail.com)) on 2017/09/28 08:05:14 +0000


------
I would also like to be able to let users disable the creation of products, however they may edit existing ones.

andris-sevcenko:
> Created by: Alan Cantes ([alan@t10interactive.co.uk](alan@t10interactive.co.uk)) on 2017/12/10 14:46:35 +0000


------
+1 to this. It&#39;d be great for a client to have a user group for product editors, allowing them admin access to create products but not publish them.

andris-sevcenko:
> Created by: Anonymous ([seskie007@gmail.com](seskie007@gmail.com)) on 2018/02/19 06:57:08 +0000


------
This feature is very important, i wanted to give user groups access to create products in commerce, and when they log in , each group type should only see their products

andrelopez:
This feature will be perfect to add support to Craft Commerce for Stripe Connect with our Stripe Payments plugin. This will help to easily associate a vendor with a product on Marketplaces.

https://github.com/enupal/stripe/issues/1

lukeholder:
@andrelopez Do you need an author/owner field on a product? Right now products have no author like entries.
Are you wanting only the creator of the product to edit/delete it?

andrelopez:
Hi @lukeholder Yes exactly, I need a way to associate a Product to a User (Vendor), we currently have `Vendors` elements which has associated a  User and `Connects` elements where you associate one or more products to one vendor. Basically a Vendor could have one or more products. One way is to have the author feature like entries, the other way that I was thinking is to add a filter to the ProductsQuery if a Vendor is logged in the CP where I can join with Connects to only show the products that are created by that Vendor. Please let me know if that makes sense

lukeholder:
- [ ] For each product type gets a new edit, create and delete permissions.
- [ ] Migration for users permissions table.
- [ ] Project config user group migration

lukeholder:
Will be fixed by https://github.com/craftcms/commerce/pull/1905

PR #1905, opened 2020-12-02
Detailed Permissions

lukeholder:
Product Permissions

Will fix: https://github.com/craftcms/commerce/issues/174

- [x] Renamed manage product type to edit product type permission.
- [x] Created create and delete product permission.
- [x] Implemented edit, create and delete production permission in action methods and templates.
- [x] Created migration for manage product type to edit, create and delete product type permission.

Split Manage Promotions

- [x] Edit Sales
- [x] Create Sales
- [x] Delete Sales
- [x] Edit Discounts
- [x] Create Discounts
- [x] Delete Discounts

Split Manage Subscriptions

- [x] Edit Subscriptions
- [x] Create Subscriptions
- [x] Delete Subscriptions

Split Manage Customer - @lukeholder deprecated

Keep Shipping and Tax as is.
Keep Manage Store as is.


MartinSchmid2409:
Will this feature also add an (editable) "author" field, just like we have it for all standard Craft entries? (Or was it already in a former update?)

sjcallender:
> No it won't - Products are not owned by an author like Entries.

I've never understood the Author field to identify the Owner of an entry. It's been helpful to identify who created the entry. In the same way, knowing who created a Product element would be extremely valuable.

mattstein:
> I've never understood the Author field to identify the Owner of an entry.

In past projects I’ve found it invaluable to have a single, clear user designated as the one responsible for an entry. Two cases:

1. Most commonly, you need to publicly expose the author with some biographical info or maybe a listing of their posts. (May not be the one who wrote or published the content, though often it is.)
2. For “live” content that needs periodic updates, any workflow/notifications can be directed at the person in charge of revisiting and updating the content.

The _Creator_ of an element seems like it’d be valuable for different reasons.

PR #1917, opened 2020-12-08
perf: purge orphaned customers using nested query

fernandez14:
### Problem experienced:

Haven't run purge in a while the number of orphaned customers can grow to the millions (I have found 4,5 million orphaned customers last time I ran this) and that scenario breaks the craft gc/run due the size of the id list parameter.

### Solution:

Proper nested query for the delete command.


lukeholder:
Thanks!

Issue #585, opened 2018-11-26
Partial Payments


lukeholder:
Allow partial payments on a completed order that is unpaid.

sjcallender:
@lukeholder Per our support conversation, here's the existing FR.

theskyfloor:
Is this not available yet in Commerce? I thought it was based on this chat from Brad in Discord:

<img width="420" alt="Screen Shot 2019-05-16 at 3 32 04 PM" src="https://user-images.githubusercontent.com/8887839/57885620-f652f900-77f0-11e9-9019-5e285dda157c.png">

lukeholder:
Only support partial refunds at the moment. On the list for Commerce 3.

samuelbirch:
Hey guys, 

Will this be included as part of 3.1, and if so when do you think it may get released?

We have a client requesting this feature :)

Cheers

ryanmasuga:
Is this possible (yet)? We are upgrading a client site from Craft 2 to Craft 3 and are weighing our options between using Craft Commerce going forward for subscriptions and partial payments (vs. a service like Snipcart).

There was a comment in Discord #commerce channel around 09/07/2020: "Yeah commerce can handle multiple transactions for an order" - but I can't find any evidence that is a fact or how that works.

Partial payments and multiple transactions sound very similar, but there is no specific mention in the [Commerce CHANGELOG](https://github.com/craftcms/commerce/blob/master/CHANGELOG.md).

samuelbirch:
I've created a PR for this feature: https://github.com/craftcms/commerce/pull/1903

lukeholder:
@samuelbirch Thanks. This will be in Commerce 4 https://github.com/craftcms/commerce/pull/1918

samuelbirch:
@lukeholder Would it be possible to add this into Commerce 3? or do you have an estimate when Commerce 4 will be released? - Cheers

ivarsbariss:
@lukeholder We also have a client that would really love to see the feature added in Commerce 3 or know the estimate when Commerce 4 will be released.

Thanks for the PR @samuelbirch  - I pulled it and was able to confirm that that's exactly what we need.

amici-infotech:
Hey @lukeholder  I have tried this with Stripe payment gateway. I can successfully pay the deposit. However, I can't pay remaining amount as that is throwing error from Stripe's endpoint. I guess because Stripe dashboard mark order succeeded. Is there any workaround for this:

[...]

PR #1918, opened 2020-12-09
Partial payment support

lukeholder:
## Feature Story

As a customer I should be able to commit to buy the items in my cart, and then make a partial payment on the order.

As a store administrator, I should be able to pull up an order in the control panel and make a partial payment on an order.

<hr>

## New order completion option

Prior to 3.3 completing an order (turning a cart into an order) was only possible by making a full payment on the order. (The payment could be an 'authorise only' transaction, which would be later 'captured' by the store administrator.)

You can now allow your customers to 'commit to buy' the order manually without needing to make a payment transaction when you turn on the new `allowCheckoutWithoutPayment` setting (default is set to false), and submitting a POST request to the `commerce/cart/complete` action.

```
        <form id="completeForm"
              method="post"
              action="">
            {{ actionInput('commerce/cart/complete') }}
            {{ redirectInput(siteUrl('/[[folderName]]/customer/order', { number: cart.number, success: 'true' })) }}
            {{ csrfInput() }}
            <button type="submit">Commit to buy</button>
        </form>
```

In addition to the above form params you can also optionally submit a `registerUserOnOrderComplete` param to enable this feature.

Just like the `commerce/payments/pay` action, in order to complete the order and checkout, the cart is required to have at least an email address set on it, and is subject to the same checkout settings rules: `allowEmptyCartOnCheckout`, `requireShippingMethodSelectionAtCheckout`, `requireBillingAddressAtCheckout`, and `requireShippingAddressAtCheckout`.

This will mark the order as complete, and take you to your redirect param. This will assign the default order status as normal, so be aware your store admins would need to confirm payment is made on the order before shipping it, which could be a change for your store administrators (although the setting is off by default).

<br>

## Partial Payments

Added a new config setting: `allowPartialPaymentOnCheckout` which allows front end customers to submit a partial payment amount (`paymentAmount`) to the `commerce/payments/pay` controller  action.

By default `allowPartialPaymentOnCheckout` is set to `false`.

 By default the full outstanding balance of the order is paid when submitting payment if the `paymentAmount` is not submitted.

The `paymentAmount` is the amount to be paid, in the cart's `paymentCurrency`.

The `paymentAmount` must be a secure hashed value (i.e`value="{{20|hash}}"`), we don't support arbitrary amounts entered by the customers directly. Example below of a select box with 50% and 100% options.

@mattstein as you know the cart/order stores everything in the base store currency, including what the outstanding balance is on the order, so to calculate the amount owing in a currency that could be different than the default, a conversion would need to be done:

```
{% set percent50 = craft.commerce.paymentCurrencies.convertCurrency(cart.outstandingBalance * 0.5, cart.currency, cart.paymentCurrency) %}

{% set percent100 = craft.commerce.paymentCurrencies.convertCurrency(cart.outstandingBalance, cart.currency, cart.paymentCurrency) %}

<select id="paymentAmount"
        name="paymentAmount">
    <option value="{{ percent50|hash }}">50%
        ({{ percent50|commerceCurrency(cart.paymentCurrency) }})
    </option>
    <option value="{{ percent100|hash }}">100%
        ({{ percent100|commerceCurrency(cart.paymentCurrency) }})
    </option>
</select>
 ```

Additional:

Backend store administrators have full access to entering any partial payment amount when making payment from the order edit page.  (the `paymentAmount` submitted from the control panel is not hashed).

We added a new function to the gateway interface `supportsPartialPayments():bool` (and added it to the base abstract gateway that all gateways should be extending), which allows a gateway to not allow partial payments. Defaults to `true`/ allow partial payments.


lukeholder:
This will fix https://github.com/craftcms/commerce/issues/585

PR #1933, opened 2020-12-17
Resave product when assigning product type to shipping and tax categories

pdaleramirez:
Fixed (#1927)


src/services/ShippingCategories.php
```diff
@@ -170,16 +170,15 @@ public function saveShippingCategory(ShippingCategory $shippingCategory, bool $r
...
-                    'elementType' => Product::class,
-                    'criteria' => [
-                        'typeId' => $oldProductTypeId,
-                        'siteId' => '*',
-                        'unique' => true,
-                        'status' => null,
-                        'enabledForSite' => false,
-                    ]
-                ]));
+                $this->_reSaveProducts($oldProductTypeId);
```
lukeholder, line 173: rename to `_resaveProducts`

pdaleramirez: Rename done

PR #1939, opened 2020-12-24
orders/purchasables-table fix

samuelbirch:
The add line item feature in the order edit page breaks if there are any soft deleted purchasables.

This fix excludes any soft deleted elements from the results.

It also orders the results by purchasable description.


nfourtythree:
Hi @samuelbirch 

Thank you for reporting this. It has been fixed in 9500829 and will be included in the next release.

I have created an issue for sorting the purchasbales table here: #1940

Thanks!

PR #1948, opened 2021-01-07
Include discounts in min shipping total calculation

### LINKED ISSUE #1800: Shipping rate seems incorrect when a discount is applied
**Description**
A client has let me know that when a certain discount is applied to an order, it affects the shipping rate of the order in a way that they did not expect. Please can you let me know if this can be adjusted. 

**To Reproduce**
How, and what happened?
1. Order total came to £124.99
2. Discount the customer applied was 25% off the order total, which meant the order total was adjusted to £99.99
3. The shipping method the customer chose has a conditional for free shipping on orders over £100
4. This order qualified for the free shipping as it didn't seem to factor discount applied

**Expected behavior**
What did you expect to happen?
I would expect the shipping rate to be taken from the order total, after discount adjustments have been made, but it doesn't seem that way. Is there any way to adjust this?

**Additional info**
- Craft version: 3.4.27
- Commerce version: 3.1.4

lukeholder:
Fixes #1800

Needing to add an option to opt into this behaviour since it is a breaking change.

- [x] new column on shipping rule `minMaxTotalType`
- [x] Add the above column as a dropdown next to minTotal in shipping edit page

 Once that is done we can merge.


nfourtythree:
To get this early, change your `craftcms/commerce` requirement in you project's `composer.json` to:

```json
"require": {
  "craftcms/commerce": "dev-feature/3.x-shipping-total-with-discounts#b56aaecdaf814d62d231c5886ad0a621aa0355b8 as 3.2.13.2",
  "...": "..."
}
```

Then run `composer update`.

pdaleramirez (approved):

PR #2000, opened 2021-02-13
Order Notices

lukeholder:
Orders now have notices. Notices are added to an order during recalculation and/or validation to notify the customer/user something has changed in the order.

Currently the following notices are created:
- Previously valid coupon on the order has been removed form the order. The reason it is not valid is explained in the notice.
- Previously valid shipping method has been removed from the order.
- A line item's purchasable that is no longer available would previously silently be removed from the cart, now the customer is notified the item was removed from the cart.
- If the sale price of a line item changes (sale no longer applies for example) the customer is notified.

In the future we can add additional notifications for things like:
- A coupon was automatically added to your cart
- Free items were added to your cart automatically since you qualify for a discount/promotion
- etc

Notices are accessed and work exactly like model errors.

Twig:

`order.getNotices()`. returns an array of all notices.
`order.getNotices('couponCode')`. one can pass an attribute to only get notices related that attribute.
`order.getFirstNotice('couponCode')`
`order.getNoticeSummary(true)`. Returns the notices for all attributes as a one-dimensional array. If set to true every notice message for each attribute will be shown otherwise only the first notice message for each attribute will be shown.
See src/elements/traits/OrderNoticesTrait.php

Clearing notices:

Notices remain on an order unless they are cleared. To dismiss notices on the cart the following are both possible:

Controller cart update action:

```twig
            <form method="post" class="js-form">
                {{ actionInput('commerce/cart/update-cart') }}
                {{ hiddenInput('successMessage', 'Coupon notice dismissed'|hash) }}
                {{ hiddenInput('clearNotices[]', 'couponCode') }}
                {{ csrfInput() }}
                <input class="" type="submit" value="Dismiss"/>
            </form>
```

```twig
            <form method="post" class="js-form">
                {{ actionInput('commerce/cart/update-cart') }}
                {{ hiddenInput('successMessage', 'All notices dismissed'|hash) }}
                {{ hiddenInput('clearNotices') }}
                {{ csrfInput() }}
                <input class="" type="submit" value="Dismiss"/>
            </form>
```


lukeholder:
@nfourtythree need your help adding notices to the order edit page.

nfourtythree (changes requested):
Just noticed a couple of things, happy to chat!

PR #2027, opened 2021-03-01
Fixed totalShippingCost attribute not returning a value

pdaleramirez:
Fixed a bug when getting the totalShippingCost attribute not returning amount value.


pdaleramirez:
Please ignore latest fix, I will revert it and change it to a migration instead.

nfourtythree (approved):
Looks good to me!

PR #2032, opened 2021-03-02
cp orders - allow purchasables that are enabled globally

samuelbirch:
This will check for a cp request and if so will then use the product->enabled property instead of enabledForSite.

This means that purchasables can be disabled for all sites (and therefore not show on the frontend) but still available to purchase via a cp order.


lukeholder:
3.3.1 will have a `purchasableAvailable` event you can hook into:

```php
Event::on(
    Purchasables::class,
    Purchasables::EVENT_PURCHASABLE_AVAILABLE,
    function(PurchasableAvailableEvent $event) {
        if($order && $user = $order->getUser()){
            $event->isAvailable = $event->isAvailable && !$user->isInGroup(1); // Group ID 1 not allowed to have purchasable in the cart.
        }
    }
);
```

Due out tomorrow.

PR #2036, opened 2021-03-05
Fixes long order references on order detail page

samuelbirch:
If an order has a long reference, then it breaks out of the containing meta box.

This fix applies the same styling as the order number.


nfourtythree:
Hi @samuelbirch 

I am not too sure what is going on here but your PR is showing as having no changed files?

I went ahead and fixed what I believe you were getting at in https://github.com/craftcms/commerce/commit/aff99161f210efbc5ed2ba28295e089702470818

This will be included in the next release of Commerce.

Thanks!

samuelbirch:
@nfourtythree Thanks, my mistake it looks like i did a PR on the wrong branch.

PR #2044, opened 2021-03-10
Fixed missing `error` and `success` keys on complete payment

### LINKED ISSUE #2043: Complete payment action does not return appropriate data for success and failure when submitting via ajax
**Description**
On successful payment completion, an `success` key with the value of `true` should be included in the JSON response when submitting via AJAX.

On unsuccessful payment completion, an `error` key with a resulting error messaged should be included in the JSON response when submitting via AJAX.

nfourtythree:
Fixes #2043 

Can you see anything wrong with this @lukeholder 

My only other note is that I didn't use `asErrorJson` because that only passes back and error and the cancel URL was already being returned.


lukeholder:
Looks fine.

PR #2049, opened 2021-03-17
Order Notices

lukeholder:
### Description

As a customer, I would like to be notified during checkout that the order have changed in ways I was not directly responsible for.

For example, when coming back to my order, and finishing checkout, if the price of a product has changed I should be able to clearly see that before checkout.

As a developer, I would like to be able to show my customers things that change in the cart while they check out. I would like to display these changes by the type of change, and know what part of the order changed.


mattstein:
> @mattstein No longer support cleaning notices by attribute, only a clear all for now. Will look at adding support for additional options on the front end in the future.

Couldn’t I still `{% do cart.clearNotices(null, 'couponCode') %}`?

mattstein:
**OR** you can ignore that last tedious comment if these changes are 100% accurate:
https://github.com/craftcms/docs/commit/93f4feb0836e9bc2a0807ae1787922e47f16cccf

lukeholder:
All these doc changes (craftcms/docs@93f4feb) look correct to me @mattstein, thanks!!!

Issue #220, opened 2018-03-28
Excluding User Groups from Sales/Discounts


andris-sevcenko:
> Created by: Ahmad Dowla ([adowla@bespokeweb.nyc](adowla@bespokeweb.nyc)) on 2016/03/14 16:05:33 +0000
Votes at time of UserVoice import: 3

------
You can currently limit sales/discounts to specific user groups. What I'd like is the ability to exclude user groups.

Here's a use case: I have a client who does wholesale pricing at 50% off most products. In Craft Commerce, I would do this through a perpetual sale limited to a 'Wholesaler' user group. But if I want to simultaneously run a sale for regular customers--say a sitewide 15% discount--folks in the 'Wholesaler' user group would get a discount that's not exactly 50% (either 15% off 50% or 50% off 85%) because the sales would overlap.

Ideally, I would be able to create a sale that applies only to guests and users under a 'Customer' user group but not to users under the 'Wholesale' group.

PR #2057, opened 2021-03-26
Add new discount promotions user group excluded condition

lukeholder:
### Description

Removed the All Groups condition and replaced it with 4 user condition options:

- All customers (users or guests) (ANY_OR_NONE)
- Discount applies to logged in users that must be in all of these groups: (INCLUDE_ALL)
- Discount applies to logged in users in at least one of these groups: (INCLUDE_ANY)
- Discount available to any logged in user not in these groups: (EXCLUDED)

### Related issues
#220


mattstein:
@lukeholder I see you requested review and then turned this back into a draft. Does that mean “review but don’t merge when finished”?

mattstein:
FYI, alphabetized translation strings which made a gnarly merge conflict so I force-pushed to delete that commit.

src/templates/promotions/discounts/_edit.html
```diff
@@ -185,19 +185,38 @@
...
-                name: 'groups',
-                options: groups,
-                values: discount.getUserGroupIds(),
-                errors: discount.getErrors('groups'),
-                class: 'selectize fullwidth',
-            }) }}
-        {% endif %}
+        {{ forms.field({
+            id: 'user-condition',
+            label: 'User Condition'|t('commerce'),
```
mattstein, line 215: Would this make more sense as “User Group Condition” rather than “User Condition”, since every available option relates to user _groups_ rather than individual users?

src/templates/promotions/discounts/_edit.html
```diff
@@ -185,19 +185,38 @@
...
+            instructions: 'Choose a condition for user restriction.'|t('commerce'),
```
mattstein, line 216: If it makes sense, I’d propose changing this to:

> Choose which user groups should match this discount.

That better frames “Any or none” since it will apply by default and not have any limiting impact, and makes it easier to consider the additional options.

src/models/Discount.php
```diff
@@ -142,11 +142,12 @@ class Discount extends Model
...
      */
     public $hasFreeShippingForOrder;
-
+    
     /**
-     * @var bool Match all user groups.
+     * @var string User condition type.
      */
-    public $allGroups;
+    public $userCondition;
```
mattstein, line 149: Being picky, but maybe this would be more accurate as `$userGroupCondition` with a more explanatory description like “Type of user group condition that should match the discount. (See `getUserConditions()`.)”

src/models/Discount.php
```diff
@@ -341,6 +342,16 @@ public function getPercentDiscountAsPercent(): string
 
         return Craft::$app->formatter->asPercent(0);
     }
+    
+    public function getUserConditions(): array
+    {
+        return [
+          DiscountRecord::CONDITION_USERS_ANY_OR_NONE => Craft::t('commerce', 'Any or none'),
+          DiscountRecord::CONDITION_USERS_INCLUDE_ALL => Craft::t('commerce', 'Include all'),
```
mattstein, line 350: Consider changing to `Include all…` to indicate reliance on additional information.

src/models/Discount.php
```diff
@@ -341,6 +342,16 @@ public function getPercentDiscountAsPercent(): string
...
+          DiscountRecord::CONDITION_USERS_INCLUDE_ANY => Craft::t('commerce', 'Include any'),
```
mattstein, line 351: Consider changing to `Include any…` to indicate reliance on additional information.

src/models/Discount.php
```diff
@@ -341,6 +342,16 @@ public function getPercentDiscountAsPercent(): string
...
+          DiscountRecord::CONDITION_USERS_EXCLUDE => Craft::t('commerce', 'Exclude')
```
mattstein, line 352: Consider changing to `Exclude…` to indicate reliance on additional information.

mattstein (changes requested):
I committed fixes to string casing and suggested changes in individual comments. Let me know if anything doesn’t make sense!

lukeholder:
Fixes https://github.com/craftcms/commerce/issues/220

Issue #2106, opened 2021-04-26
Add an on check availability event


samhibberd:
**What are you trying to do?**
Control whether an item is available based on some custom logic, in our case the value of custom fields eg custom date fields to stopSellingOn, startSellingOn 

**What's your proposed solution?**
Add an ON CHECK AVAILABILITY event to both product and purchasable `getIsAvailable()` methods. 

Added side effect is that these purchasables will then get picked up by the new notices logic and removed on recalculation!!

nfourtythree:
Hi @samhibberd 

This has been added to the upcoming `3.3.1` release due soon.

You can track the status of `3.3.1` here: https://github.com/craftcms/commerce/pull/2103

Thanks!

PR #2107, opened 2021-04-27
Add Check Variant Availability Event

samhibberd:
Implements the event suggested here: #2106

Allows plugins and modules to control whether an item is available based on some custom logic eg think custom variant date fields stopSellingOn startSellingOn

By adding this event to the variant getIsAvailable() method you have the flexibility of being able to add custom logic to determine if a product / variant is available for purchase. Also great when combined with the new cart notices feature as it allows you to ensure variants are picked up by the recalculation and the correct notice is added when / if the variant becomes unavailable.

Cleanly replaces a lot of custom behavior logic / template that would be required to get something like this working.

### Related issues

#2106


lukeholder:
Thanks so much for the PR, but we have implemented this feature in the following PR is a slightly different way: https://github.com/craftcms/commerce/pull/2109

Should be out next week.

samhibberd:
Awesome! Obviously your way is much better!!! Perfect for us.

Thanks for jumping on this sooo quickly 🎉

PR #2119, opened 2021-05-03
Don't log an error when deliberately canceling sending order mail

boboldehampsink:
### Description

Cancelling an order mail now logs an error. In my case this clogs up Sentry, while technically this is deliberate, not an error.


lukeholder:
Thanks. Accepted.

boboldehampsink:
Thanks luke

boboldehampsink:
@lukeholder looks like you only closed this but didn't merge?

lukeholder:
Sorry about that, pressed the wrong button.

Issue #2125, opened 2021-05-07
Removing qty on lineitem results in TypeError


joostwaaijer:
**Description**
When removing the qty on a lineItem and saving the cart, you'll get a TypeError Unsupported operand type.

**To Reproduce**
How, and what happened?
1. Add any product to the cart
2. Go to cart
3. Leave qty blank
4. Submit an update cart request

**Expected behavior**
LineItem should be removed when the qty value is empty or set the qty to 0.

**Additional info**
- Craft version: 3.6.13
- Commerce version: 3.3.1
- PHP version: 8.0.6
- Database driver & version: Postgres 12.6
- Other Plugins & versions:

pdaleramirez:
@joostwaaijer Thank you for reporting this bug. This bug only occurs on PHP 8. I created a PR for the fix you can track it here #2131

joostwaaijer:
Thanks!

lukeholder:
This has been fixed in 3.3.2 that has been released. Thanks.

PR #2131, opened 2021-05-10
Fixed a bug where submitting a blank line item quantity throws "Unsupported operand types: int + string" error in PHP 8

pdaleramirez:
Fixed (#2125)


lukeholder:
Thanks!

PR #2134, opened 2021-05-11
Fixed #2117 dashboard widgets showing incorrect dates

### LINKED ISSUE #2117: Dashboard Widget "Total revenue" missing orders
**Description**
When inserting the "Total revenue" widget, and selects a specific dateragen, for instance May 1 to May 2, the widget is missing ordres. The end date always says that the orders are zero, even if there are ordres. I also experience that it calculates order with, that are ordered before May 1 with.

**Expected behavior**
I expect that when i select the periode from May 1 to May 2, only the selcted dates, and the dates in between is calculated.

**Additional info**
- Craft version: 3.6.12
- Commerce version: 3.2.17.4
- PHP version: 7.4

nfourtythree:
### Description
Dashboard widgets were not showing the correct dates on charts based on a timezone mismatch.


lukeholder:
merged stat tests branch into this one for now.

PR #2140, opened 2021-05-17
Longer debounce time and cancel previous requests when searching customers on the order edit page

### LINKED ISSUE #2137: Customer selection/creation field behaving weirdly
**Description**
We just upgraded from Craft 3.4 and Commerce 3.19 to the latest versions of both. While testing out the order creation process, I've been running into some issues creating a new customer. 

First thing I noticed is as I type in an email address, it fires off a new ajax call to `admin/actions/commerce/orders/customer-search` for seemingly every keystroke. And each request takes longer than the one before it to return.
image
Granted, this is on our staging server, and without the DB resources we have on prod, and our DB has ~167k customers, but this still doesn't seem optimal. In our situation, it might save time if the ajax request only happened once we finished typing. 

Then, once those requests have returned, if no matching customers are found, the "Create Customer" option is not shown immediately if I had clicked out of the field while it was searching. I am just seeing the Customer field like:
image

If I then click in it once, nothing happens except I see my cursor in it (text is greyed out). If I click on it a 2nd time, _then_ I see the Create Customer option
image

I feel like that "Create Customer" option should be more readily available. It should still pop up even if I had clicked out of the field. My client didn't figure out that a 2nd click was necessary to bring that up.

**Additional info**
- Craft version: 3.6.13
- Commerce version: 3.3.1
- PHP version: 7.3
- Database driver & version: MySQL 5.7

nfourtythree:
Fixes #2137


lukeholder (approved):
Thanks

PR #2144, opened 2021-05-21
Fix donation tax bug

bighoho123:
Tax adjustment was incorrectly applied

### Description

When a tax rate is **NOT** targeting "anywhere else" and the address is not matching, there is some logic missing. 

Line 250 had the correct logic


lukeholder:
Thanks!

PR #2151, opened 2021-05-25
Add options to give finer control when calculating top product revenue

### LINKED ISSUE #1919: Top Products widget shouldn't include Tax amount in revenue calcs (and maybe not Shipping either?)
Revenue totals in the **Top Products** widget appear to be calculated using the line item `subtotal` (including tax and shipping), rather than the `total` (price × qty).

It's preferable to exclude tax amounts from this view, since taxes aren't actually considered revenue for a merchant — From an accounting view, they're a current liability, collected on behalf of some other entity.

In _many cases_, clients _may_ want to also exclude shipping amounts from the Top Products calculations, to get a more pure view of sales volumes.

In the current schema, it's probably not trivial to extract the tax and shipping costs from a line item individually, so...

**Proposed interim solution:** Expand the Type options in the Top Products widget settings:
* Sales Revenue (without Tax/Shipping)
* Total Collected (with Tax/Shipping)
* Quantity Sold

Eventually — as schema updates allow — excluding Tax amounts should probably be the default. 🙏🏼

nfourtythree:
### Description
This introduces "revenue options" when calculating top product revenue. It gives the ability to remove things like line item shipping and extra taxes from the calculations.

This closes #1919 

### TODO
- [x] Wording review (most notably the instructions text)
- [x] Translations


nfourtythree:
@mattstein After our conversation the other day it would be good to get your input on the wording on this one.

mattstein:
@nfourtythree After some staring and typing I arrived at this:

Screen Shot 2021-05-25 at 11 36 09 AM@2x

Nitpick detail:

1. Capitalize `o` in `options` for consistency with other labels.
2. Adjust field description for brevity, and to include “product revenue total” so it doesn’t have to be repeated for each option (wasting bytes and brain-parsing).
3. Edit down each checkbox field description in conjunction with the previous item.
4. Use possibly-objectionable phrase “built-in” to describe inclusive tax and avoid the tricky phrase “Include inclusive”.
5. Edit Tax description to reinforce distinction between it and inclusive tax.

I think this is easier to read and more clear, but take or leave whatever you’d like!

nfourtythree:
Thanks @mattstein have pushed an update with wording in line with what you said.

Slight tweak I added "when" to the "Revenue Options" instructions as I felt it needed it.

It is just the "built-in" and "separate" that I am a little unsure on. We don't really use those words when describing those taxes anywhere else. But it does make sense to me.

Thinking maybe @lukeholder might have some input here.

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+### Added
```
lukeholder, line 5: Move this to CHANGELOG-3.4.md

lukeholder (changes requested):

mattstein:
Ticked _Wording review_ since that’s very much happened, and _Translations_ since I assume that means me reviewing them.

It looks like @nfourtythree still needs to respond to @lukeholder’s request for changes so I’ll _not_ merge this as my last step.

mattstein (approved):

src/stats/TopProducts.php
```diff
@@ -22,6 +23,22 @@
  */
 class TopProducts extends Stat
 {
+    /**
+     * Constants for the type of stat query.
+     * @since 3.x
```
brandonkelly, line 28: @nfourtythree docblocks only apply to the thing that is defined directly after them. You can’t have one that acts as a heading for multiple things that follow it. So in this case `TYPE_QTY` is the only one that gets a description of “Costants for the type of stat query”, and “Since 3.x”.

nfourtythree: Good catch @brandonkelly these were placeholder for me to flesh out later and turns out I forgot they were there 🤦 

Will fix on the `3.4` branch

PR #2185, opened 2021-06-03
Improve error explanation for discount `perEmailLimit` when an email is `null`

### LINKED ISSUE #1455: Per Email Address Discount Limited coupons before checkout
**Description**
Not sure if this a bug or expected behaviour but I'm trying to give email newsletter signups a 'Per Email Address Discount Limited' single use coupon to use on their first purchase. This means they are always a guest to start with before checkout. 
If you try to use a single use coupon without an email address (e.g. in cart) the coupon shows as 'This coupon is limited to 1 uses'. The discount only works if they proceed to checkout and use an email address that's not already had the discount and then apply the coupon. It seems the system returns 'discountPerEmailLimitValid' false if there's no email address in the cart (line 960 in commerce\services\Discounts.php).

**Expected behaviour**
I expected the discount to always work for guests so their cart reflects the correct price before proceeding to checkout where they apply their email address and details. I know I expect that as a user on other sites and often check that a coupon is valid and gives me discount before proceeding with filling out forms and handing over my data.

**Additional info**
- Craft version: Craft Pro 3.4.20
- PHP version: 7.4.1
- Database driver & version: MySQL 5.5.5
- Plugins & versions: Commerce 3.1.4

**Additional context**
Really I'm using this as a work around as I'd like to use unique coupon per new customer but there's no easy way to bulk create coupons (and use them in Mailchimp) at this point without a lot of custom API faffing.

nfourtythree:
### Description
Closes #1455


lukeholder (approved):

mattstein (approved):

mattstein:
Handled the changelog merge conflict because it seemed tough to screw up, and merging now since I was only reviewing translation.

PR #2190, opened 2021-06-07
Ability to download a collated/merged PDF from the order index

### LINKED ISSUE #1785: Option to download order pdfs as one combined pdf instead of zip
**What are you trying to do?**
Downloading multiple order pdfs which now generates an zip file

**What's your proposed solution?**
Add an option to instead get a combined pdf

**Additional context**
This would be handy for printing a package slip. One pdf so only hit the print button once instead of opening multiple pdfs from the zip file

lukeholder:
### Description

As a store admin, I want to download PDFs for a collection of orders from the order index page as a single PDF file. Currently I am able to only download multiple PDFs as separate files inside a zip.

CleanShot 2021-06-07 at 19 09 10@2x

Fixes #1785


src/translations/en/commerce.php
```diff
@@ -383,7 +385,7 @@
     'Don’t apply any subsequent discounts to an order if this discount is applied' => 'Don’t apply any subsequent discounts to an order if this discount is applied',
     'Don’t apply this tax when the address has a valid VAT ID' => 'Don’t apply this tax when the address has a valid VAT ID',
     'Download PDF' => 'Download PDF',
-    'Download as Zip file' => 'Download as Zip file',
+    'Download PDF…' => 'Download PDF…',
```
mattstein, line 388: Is it important to have `Download PDF` _and_ `Download PDF…` strings?

lukeholder: Yeah it is weird, one `Download PDF` is used for a button with a dropdown arrrow next to it, and the other  `Download PDF…` acts like the export HUD on the element index, so I used the same convention there.

PR #2191, opened 2021-06-07
Various HTML improvements.

Kravimir:
Changed an incorrectly used &lt;label&gt; to a &lt;legend&gt; with a new &lt;fieldset&gt; element.

Added a few missing &lt;label&gt; elements. (The use of a placeholder attribute does not negate the need for a &lt;label&gt; element.)

Added a &lt;span&gt; to wrap the text content of more &lt;label&gt;s and added a label class to such new and existing &lt;span&gt; elements.

Added an &lt;abbr&gt; element for the "Qty" table columns.

Added an appropriate scope attribute to &lt;th&gt; elements.

Replaced most &lt;br&gt; elements with one of two classes ("block" and "mb-line").

Corrected a typo in an existing aria-label attribute.


nfourtythree:
Hi @Kravimir 

Thank you for your PR. Just a couple of questions.

> Added a <span> to wrap the text content of more <label>s and added a label class to such new and existing <span> elements.

What is the purpose behind this change? I couldn't see a `.label` class and it seems, on the face of it, that adding an extra element is not required.

> Replaced most <br> elements with one of two classes ("block" and "mb-line").

I see that you added some CSS rules for elements with `.block` the `block` class is provided by tailwind so I am not sure that those rules are needed. I feel like your addition of `.mb-line` could be removed and use `mb-5` as provided by Tailwind.

Looking forward to hearing your thoughts, and thanks again.

Kravimir:
Thanks for responding promptly and for asking good clarifying questions.

Ah. You expect people to use Tailwind. I prefer the flexibility and lack of overhead of not using any CSS frameworks.

Yes, I didn't add the "label" class to the stylesheet. It's there as a style hook for people using custom styles (without having to incur the maintenance overhead of using a custom template). The extra element is useful in some cases (especially for checkboxes and radio buttons), and for these templates, it improves consistency. Also using a class as the key selector performs much better than an element does.

REMs and EMs aren't always equal, but in this case I suppose it doesn't matter much. I would be happy to revert the stylesheet changes and swap "mb-5" in for "mb-line", if you would prefer.

nfourtythree:
> Ah. You expect people to use Tailwind. I prefer the flexibility and lack of overhead of not using any CSS frameworks.

The example templates use Tailwind by default as you can see here: https://github.com/craftcms/commerce/blob/develop/example-templates/build/shop/_private/layouts/index.twig#L8

This means all the Tailwind classes are included and ready to use.

> Yes, I didn't add the "label" class to the stylesheet. It's there as a style hook for people using custom styles (without having to incur the maintenance overhead of using a custom template). The extra element is useful in some cases (especially for checkboxes and radio buttons), and for these templates, it improves consistency. Also using a class as the key selector performs much better than an element does.

I think adding a class that has nothing attached and an element that serves no real purpose does not add any value to the example templates. The role of the example templates, although opinionated, is designed to be a slim representation of how the functionality of Commerce works when it comes to front end implementation.

We recommend that people use these templates to learn and as a reference when building a project using Commerce. They are a great base to get started with but we generally expect that people will take the code and adapt it to their project, likely using their own CSS and restructuring the HTML as required.

If adding the spans is something that will help speed up your building process that is great and I would recommend maintaining a fork of the repo with those changes included. Unfortunately, I don't think currently that change has a place in the core example templates.

The other changes in this PR are great (`abbr`, column scopes, aria etc) so if you would like to revert/fix up the other points we will happily accept those changes.

Thanks for your time, work and communication on this!

Kravimir:
I didn't think to look for external stylesheets that weren't part of the repository. Oh, well; it doesn't matter now.

Is there anything else I need to do before my (remaining) changes are merged in?

lukeholder:
We will be doing a new review of the example templates with our in house accessibility engineer, and will start a new PR with it's own goals. Thanks for the nudge.

PR #2192, opened 2021-06-08
Prevent empty `commerce.yaml` file being created on project config rebuild events

engram-design:
When running `./craft project-config/rebuild`, it'll often produce an empty `commerce.yaml` file. This is because not required, as commerce PC data is stored in its own folder, not a single YAML file.

For reference, whatever it's worth, I've added this fix on [Formie](https://github.com/verbb/formie/commit/5dfb3a43ed902258e37a8c3c7d30387eba869091) which had the same issue.

### DIFF
--- a/src/helpers/ProjectConfigData.php
+++ b/src/helpers/ProjectConfigData.php
@@ -60,7 +60,7 @@ class ProjectConfigData
             ];
         }
 
-        return $output;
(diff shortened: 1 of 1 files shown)


andris-sevcenko (approved):

lukeholder:
Fixes #2194

Issue #2213, opened 2021-06-18
Unable to save System Settings


engram-design:
I'm getting a `Couldn’t save settings.` CP error when trying to save the System Settings. No error is shown for a field.

Diving into the validation handler, this is because it gets the following error:

```
'orderPdfPath' => [
    0 => 'Order Pdf Path cannot be blank.'
]
```

As this setting has been removed from the UI (deprecated in code), it's throwing this error, but more importantly, not showing it as the setting doesn't exist in UI. It should probably be removed from the required validation rule https://github.com/craftcms/commerce/blob/develop/src/models/Settings.php#L507

Note this was a Craft 2 upgraded site, but pretty sure this would effect all installs?

**Additional info**
- Craft version: 3.6.17
- Commerce version: 3.3.4.1
- PHP version: 7.2

lukeholder:
This has been fixed for the next release including proper migration of the `orderPdfFilenameFormat` and `orderPdfPath` to the new pdf model created on migration.

brandonkelly:
Commerce 3.3.5 is out now with the fix for this. Thanks for reporting!

PR #2214, opened 2021-06-18
Unable to save System Settings due to PDF settings

lukeholder:
### Description

We should remove the required validation, but also, need to figure out the way to get the settings during the migration. I have tried here to get them from the project config directly, but that doesn't seem to work.


lukeholder:
Fixed #2213

Issue #2222, opened 2021-06-23
Rounding error in payment controller results in "Partial payment not allowed." error


joostwaaijer:
**Description**
We are experiencing some issues in a shop where we have setup various currencies for customers to pay with.
We have setup the USD as primary currency and AUD, CAD, EUR and GBP as secondary currencies. When a customer tries to pay in a secondary currency, we sometimes get rounding issues in the pay action, and then receiving the error that the payment the customer is trying to do, is a partial payment, which is not allowed. After some debugging we found out that the amount the customer wants to pay is 94.2399999 while the outstanding balance is 94.24.

I've found that the following check is blocking our customers from going to the selected payment gateway.
https://github.com/craftcms/commerce/blob/develop/src/controllers/PaymentsController.php#L430

**To Reproduce**
How, and what happened?
1. Add USD as primary currency with a rate of 1.0
2. Add any secondary currency with a rate of 0.718
3. Get a cart with a total cost of $94.24 and set the currency to the secondary one.
4. Select a gateway and perform the pay action

**Expected behavior**
The `$paymentAmountInPrimaryCurrency` should be correctly rounded of to two decimals to be the same amount as `$order->getOutstandingBalance()`

**Additional info**
- Craft version: Craft Pro 3.6.17
- Commerce version: 3.3.4.1
- PHP version:8.0.7
- Database driver & version: PostgreSQL 12.6

lukeholder:
Fixed with https://github.com/craftcms/commerce/pull/2226

joostwaaijer:
Thanks!

brandonkelly:
Commerce 3.3.5 is out now with the fix for this. Thanks for reporting!

PR #2223, opened 2021-06-24
Rounding error in payment controller results in "Partial payment not allowed." error

joostwaaijer:
### Description
We are experiencing some issues in a shop where we have setup various currencies for customers to pay with.
We have setup the USD as primary currency and AUD, CAD, EUR and GBP as secondary currencies. When a customer tries to pay in a secondary currency, we sometimes get rounding issues in the pay action, and then receiving the error that the payment the customer is trying to do, is a partial payment, which is not allowed. After some debugging we found out that the amount the customer wants to pay is 94.2399999 while the outstanding balance is 94.24.

### Related issues
https://github.com/craftcms/commerce/issues/2222

### DIFF
--- a/src/controllers/PaymentsController.php
+++ b/src/controllers/PaymentsController.php
@@ -425,7 +425,7 @@ class PaymentsController extends BaseFrontEndController
             $order->setPaymentAmount($paymentAmount);
         }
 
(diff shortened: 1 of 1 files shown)


boboldehampsink (approved):

lukeholder:
Replaces #2226

PR #2248, opened 2021-07-07
fix(gql): resolve product type id

thejoshsmith:
### Description
Fixes an issue where `productTypeId` doesn't resolve via GraphQL.

Screen Shot 2021-07-07 at 9 38 07 PM

I updated the `productResolverTest` but am not 100% how to get them running so have left it out for now.


andris-sevcenko:
Thank you!

Issue #2249, opened 2021-07-07
Coupon codes should be case-insensitive, but they're not on PostgreSQL


boboldehampsink:
**Description**
Coupon codes should be case-insensitive, but they're not on PostgreSQL.

Problem is this line:

```php
$discounts = $this->_createDiscountQuery()->andWhere(['[[discounts.code]]' => $code])->all();
```

it's case sensitive...  see https://stackoverflow.com/questions/7005302/postgresql-how-to-make-case-insensitive-query for solutions

**To Reproduce**
How, and what happened?
1. Create Coupon code
2. Try to submit it with different casing

**Expected behavior**
Should work, but doesn't.

**Additional info**
- Craft version: latest
- Commerce version: latest
- PHP version: 8.0.8
- Database driver & version: PostgreSQL
- Other Plugins & versions:

**Additional context**
Probably related to changes in https://github.com/craftcms/commerce/commit/df0a71907802665c1702b2c04f25087ff13f057e

pdaleramirez:
@boboldehampsink I've tested coupon code case sensitivity on Postgres database and wasn't able to encounter any issue. I still get valid discount despite of the case.

I checked the code and the matching of the coupon code case sensitivity is via PHP `strcasecmp` so this couldn't be a case of DB matching.

boboldehampsink:
@pdaleramirez the `strcasecmp` is only executed after the db query, which already yields no results because of case-sensitivity in postgres. What version of Postgres are you testing on?

boboldehampsink:
@pdaleramirez this was an actual issue that has been fixed in #2252

pdaleramirez:
@boboldehampsink Thank you for reporting it

PR #2252, opened 2021-07-09
Fix case-sensitive coupon codes

boboldehampsink:
### Description
Fixes case-sensitive coupon codes check by moving it to the database instead of via php.

### Related issues

#2249

### DIFF
--- a/src/services/Discounts.php
+++ b/src/services/Discounts.php
@@ -308,13 +308,23 @@ class Discounts extends Component
 
         // If the order has a coupon code let's only get discounts for that code, or discounts that do not require a code
         if ($order && $order->couponCode) {
-            $discountQuery->andWhere(
-                [
-                    'or',
-                    ['code' => null],
-                    ['ilike', 'code', $order->couponCode]
-                ]
-            );
+            if (Craft::$app->getDb()->getIsPgsql()) {
+                $discountQuery->andWhere(
+                    [
+                        'or',
+                        ['code' => null],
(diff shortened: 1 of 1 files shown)


boboldehampsink:
@lukeholder updated this to be based off of 3.4

boboldehampsink:
Please don't ignore this issue, I have this fork running in production because it's broken. cc @brandonkelly

angrybrad:
`ilike` is Postgres specific, so this would break MySQL. Seems like the correct fix for this is to normalize the coupon codes to lowercase on the initial save and have all of the subsequent checks do the same when comparing? @lukeholder @nfourtythree

boboldehampsink:
Oops, I thought it was for MySQL too. We can do a lowercase == lowercase check on the same lines tho.

boboldehampsink:
@angrybrad made the ilike conditional for postgres only and rebased on develop

angrybrad:
@boboldehampsink thanks for that. I still think normalization is the safer bet because MySQL case sensitivity can vary per collation, but will talk it over with the Commerce team.

lukeholder:
Going to go with your solution @boboldehampsink which is inline with

https://github.com/craftcms/cms/blob/develop/src/services/Users.php#L175-L200

Cutting a release shortly

boboldehampsink:
Thanks guys

lukeholder:
3.4.1 is now out with this PR included.

Issue #2505, opened 2021-07-30
Order Condition Formula on a Discount ignoring date field


simple-nicholas:
### Description
Order Condition Formula on a Discount is ignored under unclear circumstances (the Discount saves with no errors). I can't find any indications in any logs or anywhere what the issue is.

In this case, the formula is `order.eventDate < date('1-10-2021')`, where eventDate is naturally a date field. This syntax is apparently correct from everything I can find.

### Steps to reproduce

1. Add a custom date field to Orders in Commerce Settings
2. Create a Discount with an Order Condition Formula evaluating the date field
3. In the front end, add the discount to the cart with a date that violates the Order Condition Formula

### Additional info

- Craft CMS version: 3.7.7
- Craft Commerce version: 3.4.1
- PHP version: 7.4
- Database driver & version: MariaDB, 10.5.8

pdaleramirez:
@simple-nicholas We'll be releasing a fix for this issue. It will be possible to use an operator to compare custom date fields. In your case, you'll write your condition formula like this 
```
'2021-10-01' > order.eventDate|date("Y-m-d")
```
You can track the changes here https://github.com/craftcms/commerce/pull/2507

lukeholder:
Thanks this will be in the next release of commerce.

simple-nicholas:
Please reopen as #2507 has not resolved this issue.

Namely, Order Condition Formulas containing a date filter on a custom date field fails syntax validation.

The date filter and custom date fields appear to pass validation separately, but not together (eg the example from @pdaleramirez).

pdaleramirez:
@simple-nicholas Sorry, the correct syntax for the condition should be
```
'2021-10-01'|date > order.eventDate|date("Y-m-d")
```
You need to add the `|date` filter

simple-nicholas:
@pdaleramirez Your example also doesn't work, I'm still seeing a syntax validation error.

The issue appears to only occur specifically when the date filter is applied to a custom date field (eg removing the second date filter from your latest example passes validation).

Also, does the date filter work differently here? Your latest example uses an empty date filter (`|date`), which normally only results in a date string in the default format (in this case, "1 Oct 2021").

lukeholder:
Reopening and investigating.

simple-nicholas:
Hi, I've been asked to confirm that this issue is still being worked on / hasn't fallen through the cracks?

nfourtythree:
Hi @simple-nicholas 

Thank you for your message.

[...]

PR #2507, opened 2021-08-02
Whitelisted date twig filter on discount condition formula to allow condition operator on date attributes and fields.

pdaleramirez:
Fixed (#2505)


lukeholder:
Thanks

PR #2539, opened 2021-09-08
Wording cleanup

mattstein:
### Description

Minor edits for brevity and consistency. (Not sure it’s a bug fix, but you decide!)

- Expanded instances of “CP” to preferred, less-jargony “control panel”.
- Fixed contextual use of “fallback” to “fall back” (including source/left localization strings).
- Fixed lowercase “url” → “URL” used in sentence form.
- Optimized `settings.defaultView` field instruction for brevity.


mattstein:
Unless there are any changes I should make here, this is ready to go with the next translation order.

PR #2542, opened 2021-09-09
Update OrderElementTrait.php

NancyLussier:
Some customers need to search orders by id.


lukeholder:
This should be added by the base CMS, to search by element ID.

Please make a PR there on base element.

Issue #2546, opened 2021-09-11
Example template doesn't check shipping method before asking it's name.


aeu:
### Description
In the example template files, the order summary doesn't check to see if there's a shipping method before asking it's cost.

### Steps to reproduce

1. Create a store
2. Install sample templates
3. Step through checkout without picking a shipping method
4. Site crashes.

### Additional info

- Craft CMS version: Craft Pro 3.7.12
- Craft Commerce version: 3.4.2
- PHP version:7.3.25
- Database driver & version: MySQL 5.5.5
- Plugins & versions:
Amazon S3	1.2.15
Craft Commerce	3.4.2
Feed Me	4.4.0
Mandrill	1.3.1
Postie	2.4.9
Redactor	2.8.8

nfourtythree:
Hi @aeu 

I have just installed a fresh version of Craft and Commerce, used the `./craft commerce/example-templates` to copy over the example templates and was able to step through the checkout without any problems.

Where there any other steps that I would need to take to replicate the issue you had? A specific shipping method setup or any other settings? Were you using Commerce Pro or Lite?

Thanks.

aeu:
Hi @nfourtythree 

I'm using Pro, and Postie.

nfourtythree:
Hi @aeu 

Thank you for your reply. I am still unable to replicate the issue you described.

You mentioned the "site crashes" could you please send over a stack trace or error log when this happens.

This could help identify if something strange is happening.

Thanks!

aeu:
Hi @nfourtythree 

I didn't screenshot it unfortunately, the page just throws an exception, shipping not defined, or something along those lines, similar to what you get when you access a null variable.

nfourtythree:
Hi @aeu 

Thanks for that info. Unfortunately without a way of reliably replicating the issue there is not really much we can do. 

For the moment, I will close this issue. If you find a way to reliable reproduce the error or have a stack trace for it please add a comment here and I can reopen the issue. 

Thanks!

PR #2547, opened 2021-09-12
Fix for #2546.

aeu:
…ee if there's a shipping method before asking the cart for the total shipping cost.

### Description
In the example template files, the order summary doesn't check to see if the cart has a shipping method before asking it for the total shipping cost.

### Related issues
Fixes bug in example template.
https://github.com/craftcms/commerce/issues/2546


lukeholder:
Thanks for the suggestion, but we try to include all cart information fields in the examples so there is less dynamism for the developer to determine from looking at the front-end.

Issue #2540, opened 2021-09-08
Small weight value being rounded to 0


davist11:
### Description
We have some products (springs) that are really small with a weight of `0.0002` pounds. When you enter that weight into the weight field and save it, the weight value is then set to 0. It looks like the smallest value that you can enter in weight is `0.0006` which is then rounded to `0.001`. Nothing immediately stuck out to me in the code that would cause this. I also looked at the DB column and saw it was  `DECIMAL 14,4` so there should be enough room to support that number of decimals.

I wasn't sure if this was an expected minimum, and I need to hook into a before save event to do some rounding or something else was going on. Thanks in advance.

### Steps to reproduce

1. Add a weight of `0.0002` to a variant
2. Save it
3. See that the weight of the variant is now 0

### Additional info

- Craft CMS version: Craft Pro 3.6.18
- Craft Commerce version: 3.3.0.1
- PHP version: 7.4.15
- Database driver & version: MySQL 5.7.29
- Plugins & versions:

<table class="data fullwidth fixed-layout" dir="ltr">
            <tbody>
                                    <tr>
                        <th class="light">Amazon S3</th>
                        <td>1.2.11</td>
                    </tr>
                                    <tr>
                        <th class="light">Asset Rev</th>
                        <td>6.0.2</td>
                    </tr>
                                    <tr>
                        <th class="light">Avatax</th>
                        <td>2.1.6</td>
                    </tr>
                                    <tr>
                        <th class="light">Classnames</th>
                        <td>1.0.3</td>
                    </tr>
                                    <tr>
                        <th class="light">Command Palette</th>
                        <td>3.1.4</td>
                    </tr>
                                    <tr>
                        <th class="light">Craft Commerce</th>
                        <td>3.3.0.1</td>
                    </tr>
                                    <tr>
                        <th class="light">Feed Me</th>
                        <td>4.3.5.1</td>
                    </tr>
                                    <tr>
                        <th class="light">Imgix</th>
                        <td>2.1.0</td>
                    </tr>
                                    <tr>
                        <th class="light">Linkit</th>
                        <td>1.1.12.1</td>
                    </tr>
                                    <tr>
                        <th class="light">Navigation</th>
                        <td>1.4.14</td>
                    </tr>
                                    <tr>
                        <th class="light">Postie</th>
                        <td>2.4.7</td>
                    </tr>
                                    <tr>
                        <th class="light">Ray</th>
                        <td>1.1.4</td>
                    </tr>
                                    <tr>
[...]

PR #2564, opened 2021-09-23
Fixed a bug where weight value with 4 decimal places hides the value

pdaleramirez:
### Description

Fixed (#2540)

### DIFF
--- a/src/templates/products/_fields.html
+++ b/src/templates/products/_fields.html
@@ -237,12 +237,13 @@
         fieldClass: 'variant-properties-dimensions',
         errors: variant.getErrors('length')|merge(variant.getErrors('width'))|merge(variant.getErrors('height'))
     }, dimensionsInput) }}
-
+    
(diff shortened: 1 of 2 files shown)


src/templates/products/_fields.html
```diff
@@ -237,12 +237,13 @@
...
     }, dimensionsInput) }}
-
+    
+    {# Adding "+ 0" removes decimal zeros but retain precision #}
     {{ forms.textField({
         id: 'weight',
         label: 'Weight'|t('commerce'),
         name: 'weight',
-        value: (variant.weight and variant.getErrors('weight') is empty ? variant.weight|number : variant.weight),
+        value: (variant.weight and variant.getErrors('weight') is empty ? (variant.weight + 0) : variant.weight),
```
nfourtythree, line 246: This is a cool way of doing it. I think we need to add this treatment to the other fields like length, height etc.

nfourtythree (changes requested):

PR #2571, opened 2021-10-04
Env variable support for 3DS url setting

Saboteur777:
Billing detail update URL seems to support env variables, but they are not parsed during redirect, which this PR fixes.

### Description

image


lukeholder:
Thanks

PR #2572, opened 2021-10-04
Fixing a Twig Syntax Error

Saboteur777:
```
A template that extends another one cannot include content outside Twig blocks. Did you forget to put the content inside a {% block %} tag?
```

error was raised, removing that line fixes that.


nfourtythree:
@Saboteur777 Thanks for this!

PR #2594, opened 2021-10-25
Remove duplicate Product Type ID assignment

akshayagarwal:
### Description
typeId is already assigned at Line 465 and was repeated at Line 470 even though the value didn't change.

### Related issues
None


nfourtythree:
Thanks for highlighting this @akshayagarwal !

PR #2595, opened 2021-10-28
Commerce debug panel

nfourtythree:
This PR adds a Commerce debug panel to the debug toolbar in Craft.

On the front end of sites, it will show on a debug of the cart data.

In the CP, depending on which edit page you are currently on it will show a debug of the relevant model.

This should help developers understand what is happening with the data and also help with any related support requests.


src/Plugin.php
```diff
@@ -671,6 +677,35 @@ private function _registerElementExports(): void
...
+
+            $pref = Craft::$app->getRequest()->getIsCpRequest() ? 'enableDebugToolbarForCp' : 'enableDebugToolbarForSite';
+            if (!$user->getPreference($pref)) {
+                return;
+            }
+
+            $module->panels['commerce'] = new CommercePanel([
+                'id' => 'commerce',
+                'module' => $module,
+                'cart' => Plugin::getInstance()->getCarts()->getCart(),
```
lukeholder, line 704: Since carts are just for front-ends should we only get/show the cart on front-end requests?

src/helpers/DebugPanel.php
```diff
@@ -0,0 +1,108 @@
...
+
+        if (!Craft::$app->getRequest()->getIsCpRequest() || !$user || !$user->getPreference($pref) || !Craft::$app->getConfig()->getGeneral()->devMode) {
+            return;
+        }
+
+        Event::on(CommercePanel::class, CommercePanel::EVENT_AFTER_DATA_PREPARE, function(CommerceDebugPanelDataEvent $event) use ($name, $model, $prepend) {
+            $content = Craft::$app->getView()->render('@craft/commerce/views/debug/commerce/model', compact('model'));
+
+            if ($prepend) {
+                array_unshift($event->nav, $name);
```
lukeholder, line 100: Maybe use `\craft\helpers\ArrayHelper::prependOrAppend` ?

lukeholder (commented):
Looks good, just a couple notes.

src/Plugin.php
nfourtythree: Resolved with: https://github.com/craftcms/commerce/commit/e6af11b42fa624545ddceba07b2391837d3fb60f

src/helpers/DebugPanel.php
nfourtythree: Resolved with: https://github.com/craftcms/commerce/commit/769a0efd447efaf3f0093bda4c2945f3f850ab4a

PR #2618, opened 2021-11-24
Call `getIsEditable()` / `getIsDeletable()` from controllers

brandonkelly:
Controllers should be calling elements’ `getIsEditable()` and `getIsDeletable()` methods rather than doing their own permission checks directly.

## Products
- [x] Ensure `Product::isEditable()` is implemented
- [x] Ensure `Product::isDeletable()` is implemented
- [x] Call `Product::getIsEditable()` from controllers
- [x] Call `Product::getIsDeletable()` from controllers

## Subscriptions
- [x] Ensure `Subscription::isEditable()` is implemented
- [x] Call `Subscription::getIsEditable()` from controllers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,15 +3,16 @@
 ## Unreleased
 
 ### Added
-- Added `craft\commerce\elements\Product::isDeletable()`. ([#2606](https://github.com/craftcms/commerce/issues/2606))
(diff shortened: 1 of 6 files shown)


pdaleramirez (approved):
I tested all all the implementation (edit, save and refreshPayments) for enforceEditSubscriptionPermissions and all worked

PR #2645, opened 2021-12-20
Fixes lineItem & adjuster currency in CP view

samuelbirch:
Fixes lineItem & adjuster currency to match that of the order currency rather than the base currency in the CP edit view.


lukeholder:
The order currency can only be the single base currency.

If you are talking about payment currency, we don't show product prices in the backend in the payment currency, only the store currency.

Commerce 5 will have multi-currency support. 

Thanks.

PR #2650, opened 2021-12-23
Fix decimal for taxrate translations

jerome2710:
### Description
Tax rates should have a point instead of a comma as decimal. The translations are inconsistent, which leads to confusion with clients as their rates are rounded off.


nfourtythree:
Hi @jerome2710 

Thank you for your PR. This caused us to look further into the localisation of the tax rate inputs. 

We have made a number of translations changes, as well as making sure the tax rate input fields accept values based on the user's formatting locale. These changes will be in the next release of Commerce.

There have also been changes on Commerce 4 to make the entry of these numbers more straightforward for people using the Commerce control panel.

Thanks!

PR #2658, opened 2022-01-07
Added line item description to the order searchable keywords

lukeholder:


terryupton:
@lukeholder Is there any requirement or action to take in Craft/Commerce for this to take effect?
I have updated but I still can't seem to search via order line item descriptions.

nfourtythree:
Hi @terryupton 

The order elements (orders and carts) will need resaving in order to trigger the search index to kick in and update the keywords.

I would suggest just as an extra precaution I would just try it locally/on staging first just to make sure there aren't any other undesired effects that might happen.

Thanks.

PR #2707, opened 2022-02-18
Removed incorrect behavior description "Leave blank for all categories."

pdaleramirez:
The description is not the correct behavior. Now that we have the All categories switch, leaving the categories blank won't make the default behavior to All categories as the behavior


nfourtythree:
@pdaleramirez We only need to change the `en/commerce.php` translation file. The translation service will deal with the rest. 

This will have to go along with some other translation changes.

pdaleramirez:
> @pdaleramirez We only need to change the `en/commerce.php` translation file. The translation service will deal with the rest.
> 
> This will have to go along with some other translation changes.

I made the change now

PR #2736, opened 2022-03-14
Reduce Store Location heading top padding

mattstein:
### Description

Reduces the amount of padding above the _Store Location_ heading in the control panel.

Before:
before

After:
after


lukeholder:
thanks

PR #2741, opened 2022-03-22
Update related sales and discounts on product edit page using detailed promotions permissions

pdaleramirez:


lukeholder:
Thanks!

PR #2758, opened 2022-03-30
Improve Tax Rate docblocks

mattstein:
### Description

Took a stab at adding some comments where they were missing and being more specific in a few places.

Added some missing `@param` and `@return` declarations as well.


lukeholder:
Thanks!

PR #2774, opened 2022-04-19
Add $order variable to transaction event

bymayo:
### Description

It would be good to have access to the $order variable when a transaction is being made/or has been made.

### Related issues

Had an issue where we needed to adjust some prices/currencies before the transaction went through. But all prices were already converted in the $transaction variable.


bymayo:
@lukeholder No, your right! *facepalm* i tried everything to get this to work but didn't notice that method in the docs.

PR #2801, opened 2022-05-04
Create method to transfer all commerce data from one user to another 

timkelty:
This PR adds the ability to move all commerce data related to a user, to another user.

- Added `craft\commerce\services\Customers::moveCustomerDataToCustomer()`.
- Added the `commerce/move-data` command which calls the service method.


lukeholder:
@timkelty renamed to 

`Customers::transferCustomerData()` and `commerce/transfer-customer-data`

PR #2809, opened 2022-05-10
Fixed a bug when submitting a custom field on address namespace doesn't save the field.

pdaleramirez:
### Description

On the example templates when you add new address adding a custom field doesn't save the custom field to the element.
E.g.

Adding twig code 

```
    {{ input('text', 'fields[phone]', address.phone ?? '', {
      id: 'phone',
      class: ['w-full', 'border border-gray-300 hover:border-gray-500 px-4 py-2 leading-tight rounded']
    }) }}
```
 to the `templates/shop/_private/address/fields.twig` example template will not be saved to the shipping and billing address element.


nfourtythree (changes requested):
@pdaleramirez the only thing left to do here is tidy up the things that have been created in the DB during this test.

I looked at my test DB after it had finished and it looks like the field, layout tabs and layout were still in there. It is worth removing these after the test completes so we don't have any sideffects later on.

nfourtythree (approved):

PR #2823, opened 2022-05-24
Add warning when no product types exist on shipping and tax categories.

lukeholder:
CleanShot 2022-05-24 at 16 50 40@2x


lukeholder:
@mattstein Could I get feedback on the text that needs translations. Also not sure the arrow is something we do elsewhere. The arrow could also have RTL vs LTR implications yeah?

brandonkelly:
Made that change (744f19b20c6902a04244fdda3317b65142802a09).

PR #2834, opened 2022-06-01
Fix PaymentForm namespacing for Subscriptions

AugustMiller:
### Description

While trying to start a new Subscription in v4, I noticed that the POSTed data wasn't populating the `PaymentForm` model with a `paymentMethodId`, and would fail to start.

All this does is replicates the namespacing strategy used in `PaymentsController::actionPay()` by passing the subset of body params corresponding to the selected Gateway!

> 🚨 I _think_ this is an OK patch release, as the v4 migration guide makes note of the new namespacing feature. Some users’ Subscription forms _may_ break, though, if they hadn't yet patched them!

### Related issues

- Only other mention of a blank `paymentMethodId`, but for a different version: https://github.com/craftcms/commerce-stripe/issues/193


nfourtythree:
Hi @AugustMiller 

Thank you for this, will get it merged in and will be included in the next release of Commerce.

Thanks!

PR #2851, opened 2022-06-16
Add ability to see if order addresses match each other

nfourtythree:
### Description
Now that Addresses are elements Commerce requires a way to simply check if the addresses on the cart match each other as they should always be stored as separate IDs on the order. This is useful when building a checkout flow.


src/elements/Order.php
```diff
@@ -2864,6 +2872,14 @@ public function setBillingAddress(AddressElement|array|null $address): void
             throw new InvalidArgumentException('Can not set a billing address on the order that is is not owned by the order.');
         }
 
+        if ($address->id) {
```
lukeholder, line 2875: Didn't we remove this bock on our call together?

lukeholder (changes requested):

PR #2853, opened 2022-06-16
Throwing a detailed gateway error when hitting reload after submission.

pdaleramirez:
Fixed a bug where reloading after the payment form is submitted re-submits the form in payment.twig example templates.
97713bf2-9915-4f99-af78-dfd51a7a658f


lukeholder:
We need to find a way to do this server side.

PR #2883, opened 2022-07-06
Add new taxable subject 'Purchasable' which is unit price

lukeholder:
### Description

Currently the standard taxable subject is the line item subtotal which is `salePrice * qty - discounts` with the line item tax adjustment representing a X% tax on that total.

This introduces an item level tax that is based on the single unit price `salePrice - (discount/qty)` with the line item tax adjustment representing the calculated amount of tax per item * qty.


lukeholder:
Translations approved by @mattstein in slack.

PR #2892, opened 2022-07-14
Fixed getting product type id's

boboldehampsink:
### Description

It got the handles instead of the id's

Using this in fixtures results in the following error:

`Exception 'TypeError' with message 'Cannot assign string to property craft\commerce\elements\Product::$typeId of type ?int'`


nfourtythree:
Thanks @boboldehampsink , I just did this straight on `develop` as this PR branch wasn't up to date with it and therefore was failing a couple of tests.

PR #2894, opened 2022-07-14
Fix checkout thermometer width in example templates

alumpe:
### Description
Adds a more precise calculation for the checkout steps thermomether width. The thermometer bar now exactly ends in the middle of the current step.


nfourtythree:
Thanks @alumpe !

PR #2934, opened 2022-08-12
Added the ability to set a primary payment source for a customer

nfourtythree:


timkelty (commented):
@nfourtythree wondering if it makes sense to be an `Order::makePrimaryBillingAddress: bool` like there is a `makePrimaryBillingAddress`?

lukeholder:
> @nfourtythree wondering if it makes sense to be an `Order::makePrimaryPaymentSource: bool` like there is a `makePrimaryBillingAddress`?

Yes will add. Thanks!

PR #2981, opened 2022-09-25
Add inline parameter to DownloadsController

manud99:
Add `inline` parameter to `DownloadsController` to display PDFs directly in a browser tab.

### Description

During development of PDF files I find it handy to display them in a browser tab so I just have to reload the tab after changes. For this purpose I added an `inline` parameter to the response object which defaults to `false`. To enable it just add `&inline=true` to the URL of the corresponding request.


lukeholder:
thanks!

PR #2997, opened 2022-10-12
Add quantity field to add line item table on Order Edit page

nfourtythree:
### Description
This allows a store administrator to choose the quantity that is being added. This can get around an issue with adding a variant that has a "minimum quantity".

### Related issues
#2993


pdaleramirez (approved):
I've tested this and it is working as expected.

PR #3025, opened 2022-11-10
Fixed guest checkouts potentially auto setting addresses for inactive users

nfourtythree:


lukeholder:
This was correctly independently on develop.

PR #3040, opened 2022-11-22
Fix possible issues with “or not null” queries

nfourtythree:
### Description
[After the revert](https://github.com/craftcms/cms/commit/991396e42269463790a972dc86b9cf7bbb71caca) of the `Db::parseParam()` update (with accounting for `null` values) this PR tries to prevent any compatibility issues that could arise.

### Related issues
https://github.com/craftcms/cms/issues/11931


src/elements/conditions/orders/CustomerConditionRule.php
```diff
@@ -61,7 +61,13 @@ public function getExclusiveQueryParams(): array
     public function modifyQuery(ElementQueryInterface $query): void
     {
         /** @var OrderQuery $query */
-        $query->customerId($this->paramValue());
+        $paramValue = $this->paramValue();
+        ;
```
lukeholder, line 65: extra `;` ?

lukeholder (commented):
look good, just see the typo `;`

src/elements/conditions/orders/CustomerConditionRule.php
nfourtythree: Hmm thought my ecs fix removed this 🤔 .

PR #3055, opened 2023-01-03
commerce-specific site condition rule for zones

i-just:
### Description
Adds a `CommerceSiteConditionRule` which extends the `SiteConditionRule` and handles the site conditioning not based on the element's site, but the currently loaded site. 

That new condition rule is used when triggered via `ZoneAddressCondition`. 
I also added a migration so that “old” site rules in tax & shipping zones still work with this approach. 

Since I’ve already written it all before Brandon closed the related CMS issue as not planned, I am raising this PR. Feel free to use, adjust or discard it.

### Related issues
[#12491](https://github.com/craftcms/cms/issues/12491)

### DIFF
--- a/src/elements/conditions/CommerceSiteConditionRule.php
+++ b/src/elements/conditions/CommerceSiteConditionRule.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace craft\commerce\elements\conditions;
+
+use Craft;
+use craft\base\ElementInterface;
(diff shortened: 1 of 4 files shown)


lukeholder:
Thanks for the PR.

Going to try the following: https://github.com/craftcms/cms/pull/12601

If that fails I will reopen this PR.

Issue #3070, opened 2023-01-30
[4.x]: Deleting user with existing order/subscription shows success message


Romanavr:
### Description
I noticed that for some reason I can't delete a user after upgrading to commerce 4.x  even though the flash message always shows that "Users deleted" https://i.imgur.com/c7ZQEe8.png (however, user still exists after page refresh)
Then I did a little research & found that it was because the user (customer) had an existing order/subscription.
https://github.com/craftcms/commerce/blob/4.2.4/src/Plugin.php#L586
Order: https://github.com/craftcms/commerce/blob/4.2.4/src/services/Orders.php#L173
Subscription: https://github.com/craftcms/commerce/blob/4.2.4/src/services/Subscriptions.php#L408

Returning`false` here https://github.com/craftcms/cms/blob/4.3.6.1/src/services/Elements.php#L1954
And it shows us a success flash message however it's wrong https://github.com/craftcms/cms/blob/4.3.6.1/src/elements/actions/DeleteUsers.php#L173

This is a visual error for common/general user view only(`/*cp-url*/users`), if you'll try to delete a user on the detailed page(`*cp-url*/users/{id}`) you will get a fail flash message. The same behavior should be for the general user view.

### Steps to reproduce
1. Create a user and add an order/subscription to him.
2. Try to delete this user on the users page (`http://site.com/admin/users/all`)
3. The flash message shows "Users deleted" but after page refresh the user(s) is still there.

### Expected behavior
A flash failure message should appear describing why we can't delete the user(s). #3071 

### Actual behavior
It shows a successful flash message "Users deleted"

### Craft CMS version

4.3.6.1

### Craft Commerce version

4.2.4

### PHP version

8.1

### Installed plugins and versions

-

Romanavr:
Any update here?

gaelpleeroy:
Hi, I have the same problem here I can't delete my test user no matter the option I choose (transfer or delete content).

lukeholder:
@Romanavr @gaelpleeroy 

We have made the stop-gap measure in Commerce (#3071) to show an exception on the first undeletable user through the flash error. This will be in the next release.

Although not ideal, we expect Craft 5 will introduce a better way for Commerce to hook into user deletions.

To get the fix early, change your `craftcms/commerce` requirement in composer.json to:

```json
"require": {
  "craftcms/commerce": "dev-develop#151d0b250e16459a875df1654dabf73bd41cabf3 as 4.2.10",
  "...": "..."
}
```

[...]

PR #3071, opened 2023-01-30
Show fail flash message when trying to delete a user with an existing…

Romanavr:
… order/subscription

### Description
Shows a fail flash message that we can't delete such user because it have an existing order or subscription.

### Related issues
https://github.com/craftcms/commerce/issues/3070

### DIFF
--- a/src/services/Orders.php
+++ b/src/services/Orders.php
@@ -170,7 +170,7 @@ class Orders extends Component
 
         // If there are any orders, make sure that this is not allowed.
         if (Order::find()->customerId($user->id)->status(null)->exists()) {
-            $event->isValid = false;
+            throw new Exception("Unable to delete a user with an existing order, user ID “{$user->id}”");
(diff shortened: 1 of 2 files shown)


lukeholder:
Thanks for this PR. 

We have added this stop-gap measure in Commerce to show an exception on the first undeletable user through the flash error. This will be in the next release.

Although not ideal, we expect Craft 5 will introduce a better way for Commerce to hook into user deletions.

The fix will be included in the next Commerce release.

To get the fix early, change your craftcms/commerce requirement in composer.json to:

"require": {
  "craftcms/commerce": "dev-develop#151d0b250e16459a875df1654dabf73bd41cabf3 as 4.2.10",
  "...": "..."
}

Then run composer update.

nfourtythree:
Version `4.2.10` of Commerce is now out with this fix.

Thanks!

PR #3090, opened 2023-02-16
Don't attempt to claim another customer's cart

timkelty:
Here's how to reproduce the current bug:

- log in as `userA` (a user with a default payment source set)
- an empty cart will be created on login
  -  `paymentSourceId` set
  -  number will be stored in the `xxxxxxx_commerce_cart` cookie
- Delete your php session cookie manually (don't use `logout`).
- log in as `userB`
- At this point, you'll get an `invalid payment source` exception because it is trying to save the cart with the new user and the payment source doesn't belong to that user.

I don't _think_ we'd even want to allow a customer to change in this way, assuming the cart customer was an active/credentialed user.

I'm not 💯 this code is necessarily in the right place, but I think something close to this is what is needed.

https://app.bugsnag.com/pixel-and-tonic-inc/craftnet/errors/63e2930d8db91100088f535d?event_id=63ed50ac00ad321f8a250000&i=sk&m=oc

### DIFF
--- a/src/services/Carts.php
+++ b/src/services/Carts.php
@@ -190,8 +190,19 @@ class Carts extends Component
             ->status(null)
             ->one();
 
-        // If the cart is already completed or trashed, forget the cart and start again.
-        if ($cart && ($cart->isCompleted || $cart->trashed)) {
+        if (!$cart) {
+            return null;
+        }
+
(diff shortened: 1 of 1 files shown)


lukeholder:
Leaving this alone for now, if it comes back up for Craftnet we can have another discussion.

Issue #3109, opened 2023-03-10
[4.x]: "validateCartCustomFieldsOnSubmission" no longer works after updating to CraftCMS v4.4


jonleverrier:
### Description

Since updating to CraftCMS v4.4.0, when "validateCartCustomFieldsOnSubmission" is set to "true", it no longer validates custom fields on update cart in the front-end. It was working fine before updating.

### Steps to reproduce

1. Add a "plain text" field to Commerce > System Settings > Order Fields
2. Make it required
3. Make sure the custom "plain text" field is included somewhere in your checkout workflow which is wrapped in a `commerce/cart/update-cart` form
4. Browse to the step where the "plain text" field is included, leave blank and try and proceed to the next step
5. You will be able to proceed to the next step. No validation errors reported.

### Expected behavior

When custom fields are added to Order Fields and validateCartCustomFieldsOnSubmission is set to "true", on the front-end, they should show errors if for example, they are left blank. 

### Actual behavior

No validation errors are reported in the front-end when custom fields are added to Order Fields and set to required.

### Craft CMS version

v4.4.0

### Craft Commerce version

v4.2.5.1

### PHP version

8.0.21

### Operating system and version

Linux 5.15.49-linuxkit

### Database type and version

MySQL 10.4.25

### Image driver and version

Imagick 3.6.0 (ImageMagick 6.9.11-60)

### Installed plugins and versions

-

jonleverrier:
Just updated, to CraftCMS v4.4.1 and the issue is still there.

pdaleramirez:
@jonleverrier I can confirm the bug, but it seems to be a cms issue. We`ll report it to the cms.

pdaleramirez:
Created an issue in CMS here: https://github.com/craftcms/cms/issues/12884

lukeholder:
This is now fixed in Commerce 4.2.6 which is now compatible with 4.4

PR #3111, opened 2023-03-15
Cart custom field validation no longer works

pdaleramirez:
### Description

Fixed a bug where custom field validation doesn’t work when updating a cart. (#3109 3109)


nfourtythree:
Closing PR as it was fixed in: https://github.com/craftcms/commerce/pull/3113

PR #3121, opened 2023-03-20
Retry send email job

boboldehampsink:
### Description

Sending e-mail can fail when your e-mail provider is down, for example. But an order confirmation mail is important to get, and important to send without checking your CP for failed jobs all the time. I think it should retry a few times?


lukeholder:
Merging from #3140

PR #3130, opened 2023-04-04
Make it possible to re-order `LineItem`s

freder:
see https://github.com/craftcms/commerce/discussions/3129


brandonkelly:
My vote is to reject this – see https://github.com/craftcms/commerce/discussions/3129#discussioncomment-5535336.

lukeholder:
The alternative solution was given in #3129 for now. Thanks.

PR #3135, opened 2023-04-10
Craft Commerce now requires Dompdf 2.0.0 or later.

pdaleramirez:


lukeholder:
@pdaleramirez Please run `composer run fix-cs ./src` and `composer run phpstan` on commerce code before committing.

pdaleramirez:
> @pdaleramirez Please run `composer run fix-cs ./src` and `composer run phpstan` on commerce code before committing.

I ran ECS fix and PHP stan and also pushed the composer.lock to pass the dependency check.

Issue #3138, opened 2023-04-12
[4.x]: On cart update a new user is getting created when recently changed email of linked user.


wijnenmik:
### Description

When a customer / user has an active shopping cart linked and the user is changing his / her email address,  the shopping cart event is getting triggered and a new user with the old email (probably the old cart where the old email is in) is getting created. So when you want to change your email back to the old one, this is impossible because there is already an account with that email address

### Steps to reproduce

1. Create a user in craft
2. Make an active shopping cart for that user with an address linked
3. Change the email of this user
4. Trigger the _getCart function that will get the cart for the current session
5. A new user is getting created with the old email before the email update

### Expected behavior

The cart is getting updated with the new email and a new user is not getting created.

### Actual behavior

A new user is getting created with the old email that is probably still linked in the active cart

### Craft CMS version

4.4.6

### Craft Commerce version

4.2.6

### PHP version

8.1 / 8.0

pdaleramirez:
@wijnenmik I can replicate this issue. We'll look into this and give an update.

nfourtythree:
Hi @wijnenmik 

This has now been fixed (by #3180) and will be included in the next release of Commerce.

To get this early, change your `craftcms/commerce` requirement in your project's `composer.json` to:

```json
"require": {
  "craftcms/commerce": "dev-develop#25f8fee6971f0b2417e3170d46e1e193914f5a46 as 4.2.11",
  "...": "..."
}
```

Then run `composer update`.

PR #3141, opened 2023-04-13
Fixed user creation on associated existing cart

pdaleramirez:
### Description

Fixed (#3138)


lukeholder:
This doesn't seem like the right thing to do. We need to discuss.

nfourtythree:
This was fixed by #3180

PR #3144, opened 2023-04-19
Fix German translation

MoritzLost:
Fixes one German translation.

`wiedergeben` means to play back, which doesn't make sense in this context.


brandonkelly:
Thanks! I’ve just pulled your change into Commerce’s [Crowdin project](https://crowdin.com/project/craft-commerce). If you have any additional edits you want to make in the future, you might find it easier to submit them directly from there :)

MoritzLost:
@brandonkelly Thanks! Yeah I was too lazy for Crowdin this time 😇

PR #3171, opened 2023-05-15
Fixed a bug where hasUnlimitedStock sets to NULL if no value is passed.

pdaleramirez:
### Description
This is a fix for feed me bug https://github.com/craftcms/feed-me/issues/636


lukeholder:
This is not an issue in Commerce 4 as the default is set and the column no longer allows null.

Issue #3176, opened 2023-05-27
[4.x]: User group conditions - should migrate to the new condition builder, no obvious warnings shown


bossanova808:
### What happened?

I think user group conditions on discounts should not just be silently dropped during a Commerce 3 -> 4 migration, but a condition should be created in the condition builder:

![image](https://github.com/craftcms/commerce/assets/731309/5ef302dd-897d-4be6-b73c-eb16fb119114)

To:

![image](https://github.com/craftcms/commerce/assets/731309/607a3e73-e754-42be-a51d-2206ce0210bb)

If this is too difficult to implement then at the very least there should be clear warnings raised during the migration, and this should be noted in the upgrade documentation.

This is very important and should be solved as a priority - discounts very quickly have significant financial implications and with the current migration, the conditions are stripped with no warning, meaning these discounts then apply to all users, which is of course very much not correct...

### Craft CMS version

4

### Craft Commerce version

4

### Installed plugins and versions

-

lukeholder:
If it's not being migrated to the conditioner, then that's a bug as it should be migrated to condition builder. I will look into it. We wouldn't drop it silently on purpose.

bossanova808:
Thanks Luke.

They weren't for me on my last test migration.  Am planning on doing a new test migration early next week, so can report back then again.

pdaleramirez:
I have managed to fix this bug. You can track the PR #3177

bossanova808:
Great & many thanks - will hopefully have time tomorrow for the next test migration, so good timing.

nfourtythree:
Version `4.2.10` of Commerce is now out with this fix.

Thanks!

bossanova808:
Unfortunately, this does not work & breaks my migration

```
*** applying m220304_094835_discount_conditions
    > update in {{%commerce_discounts}} ... done (time: 0.007s)
    > update in {{%commerce_discounts}} ... done (time: 0.000s)
    > update in {{%commerce_discounts}} ... done (time: 0.000s)
    > update in {{%commerce_discounts}} ... done (time: 0.001s)
    > update in {{%commerce_discounts}} ... done (time: 0.000s)
    > update in {{%commerce_discounts}} ... done (time: 0.000s)
    > update in {{%commerce_discounts}} ... done (time: 0.000s)
    > update in {{%commerce_discounts}} ... done (time: 0.000s)
    > update in {{%commerce_discounts}} ... done (time: 0.000s)
[...]

PR #3177, opened 2023-05-29
User group condition v3 to v4 upgrade not migrating

pdaleramirez:
### Description

Fixed a bug where where user group condition values are not migration from v3 to v4.
#3176


pdaleramirez:
I checked, and it is not possible to get the user group data from the previous version if Commerce has already been upgraded to v4.

lukeholder:
Thanks Dale, I just made a tweak to continue using the discount user condition.

Issue #3178, opened 2023-05-30
[4.x]: Add ability to sync address book changes to carts using them as source addresses 


nfourtythree:
### Description
If a customer has created a cart by using an address or addresses from their address book it is reasonable to think that changes made to these addresses would also be applicable to the cart.

We want to ensure we do not change any completed orders. We need to keep the separation that currently exists in terms of the address model attached to the cart separate from the address owned by the order. This is to make sure that features regarding order editing and historic order data (post-order complete) continue to work as excepted.

Due to the nature of this, we may need to put this under a setting for the time being to avoid any unexpected behavioural changes. Something along the lines of `syncAddressBookChanges`. The setting being `true` would likely become the default in the next major release.

bossanova808:
I think there are two key functional issues to address.  These of course only apply to users with accounts who are logged in (for us this is about 80% of our orders).

1. Logged in users, using saved addresses, expect changes to those to be propagated back to their address book.  This is the behaviour of just about every shopping cart on the planet, and the natural expectation of the user when editing an existing thing (in just about any computing context).  If you do want a new address created from a save one, then a site could offer a 'save as' type option.

2. Logged in users (and new users who choose to register an account at checkout) - expect new addresses they use to be saved for re-use (that, and order history, are really the two reasons to ever both with having accounts on e-commerce sites).  Again, this should be the default - with an opt-out for those that don't want addresses saved (or, in our case, we offer a delete button on the addresses when they are shown later, to make this simple and reduce the checkout question fatigue).

These really should be the default behaviour as it matches the established patterns of online shopping, and indeed the historically established patterns of Commerce itself.  

There _may_ be rare scenarios where those behaviours are not desired (although honestly, I am struggling to think of any such case - e.g. two people ordering with the same address but different name -> they would just create a new address with the other name, right?).  I guess an opt out system might make sense, if you care to implement and support that.  But neither of these things should require custom module code, as they currently now do.

nfourtythree:
Commerce's solution for this have now been merged in ready for the next release. 

More discussion on this topic is on the pull request: https://github.com/craftcms/commerce/pull/3182

Along with all of this we have added and issue to the docs to look into improving the documentation helping developers create a coherent and robust checkout when building with Commerce (https://github.com/craftcms/docs/issues/509).

Thanks!

nfourtythree:
As an update to close this issue, these changes were released in version `4.3.0` of Commerce.

Thanks!

PR #3182, opened 2023-06-01
Sync changes to address book address to applicable carts

nfourtythree:
### Description
~By setting the `syncAddressBookChanges` config setting to `true`, when a user saves an address in their address book that is being used as a source address for a cart (`sourceBillingAddressId`, `sourceShippingAddressId`) the respective addresses will be updated on the cart.~

~This new setting is `false` by default to preserve the current behaviour in Commerce 4 and not cause any issues for developers/project owners when updating.~

When updating an address in a user's address book, if that user's address is used as a source address on any carts. These carts should be updated with the new address details.

### Related issues
#3178


bossanova808:
I may well be misreading this, but this feels a little bit the wrong way around to me.  The idea is, if a user changes an address they are using for a cart, that change is then automatically propagated back to the address book.   That is because, naturally, people will load and use (and edit) addresses, during checkout, as the most common flow (& I am currently implementing this with a call to save-address).

 Here it seems like you're implementing something else - if they change an address in their address book, whilst it is set on the cart, it should change there too.  I suppose that is true, too - albeit it will be a relatively unusual flow.

nfourtythree:
Hi @bossanova808 

Thank you for your feedback. We are trying to cover the many use cases people have for Commerce. Address books, custom addresses directly on the cart etc.

With this fix, we have hit a spot where those things are all now possible. With users and their address books, this idea is based on how places like Amazon and Shopify deal with their addresses. You are selecting an address from your address book to use on the cart and if you want to update that address, it should happen within the address book system.

If you were to pull down a fresh install of Commerce 4 and take a look at the updated example templates from this PR I think things would start to make more sense. We need to get this information and explanation into the documentation (and examples in there) to do a better job of educating developers on what is possible.

We have definitely taken your point on board and it has caused us to have discussions and make notes internally on what else may or may not be required regarding addresses during checkout in the future.

Thanks!

bossanova808:
Thanks for the extra info.  

Stupid question perhaps, but what is actually the best/recommended  way to get a vanilla Commerce install going these days - the Craft starter project doesn't include Commerce, right?  Should one use that and then manually add Commerce, or is there a more direct way?    (Be nice is there was a Commerce included version, but I don't think that exists?)

nfourtythree:
@bossanova808 the steps you have outlined with the Craft starter will probably be the quickest way currently. A suggestion of a Commerce start is a good idea, will take this back to the team to see if we can add it to the road map.

Thanks!

PR #3184, opened 2023-06-02
Fix version number in composer.json for 3.4.22

MatMc:
composer.json for v3.4.22 is incorrectly set as v3.4.21 resulting in Craft's plugin update page prompting to update even when the latest version is installed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
   },
   "extra": {
     "handle": "commerce",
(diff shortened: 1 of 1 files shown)


nfourtythree:
Hi @MatMc 

Thank you for this, will merge this in and get another release out with an update version number to make sure that all files and tags are matched up.

Thanks!

PR #3201, opened 2023-06-22
Deprecate `email` as an attribute on the `Order` element

nfourtythree:
### Description
Currently Commerce is needlessly duplicating data into the `email` column for the order element. As all customers are now user elements (either credentialed or inactive) the email for the order can be taken directly from this.

This is the first part of a two step process to remove `email` as an actual attribute on the `Order` element. In this step it is simply to tidy up the code around emails and deprecate the `setEmail()` on the element. The second part, likely to be in Commerce 5, will be to remove the `email` column entirely from the orders database table.

With `setEmail()` now being deprecated it is advised to switch any code using the method to use `setCustomer` along with the `ensureUserByEmail()` method from Craft's user service.

### DIFF
--- a/src/elements/Order.php
+++ b/src/elements/Order.php
@@ -1110,20 +1110,6 @@ class Order extends Element
     private ?string $_paymentCurrency = null;
 
     /**
-     * @var string|null
-     * @see Order::setEmail() To set the order email
-     * @see Order::getEmail() To get the email
-     * ---
-     * ```php
-     * echo $order->email;
-     * ```
-     * ```twig
-     * {{ order.email }}
-     * ```
-     */
-    private ?string $_email = null;
-
-    /**
      * @var Transaction[]|null
      * @see Order::getTransactions()
      * ---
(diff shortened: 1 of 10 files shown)


src/elements/Order.php
```diff
@@ -2316,37 +2298,37 @@ public function getUser(): ?User
...
      * @throws Exception
+     * @deprecated in 4.3.0. Use [[setCustomer()]] instead.
      */
     public function setEmail(?string $email): void
     {
+        Craft::$app->getDeprecator()->log(__METHOD__, '`Order::setEmail()` has been deprecated use `Order::setCustomer()` instead.');
         if (!$email) {
             $this->_customer = null;
             $this->_customerId = null;
-            $this->_email = null;
```
lukeholder, line 2325: any reason not to clear the `this->_email` ?

nfourtythree: Where have you seen that, I can't see it in the code?

PR #3231, opened 2023-07-25
Variable name typos

AugustMiller:
`$currentPlan` had an extra `t` at the end in a couple of instances!

~~This is an inconsequential patch—in both cases, the variable wasn't actually used in-scope… and to my knowledge, interfaces aren't required to have the same argument names (just the types).~~

See below! 😬

🌱

### DIFF
--- a/src/base/PlanInterface.php
+++ b/src/base/PlanInterface.php
@@ -21,5 +21,5 @@ interface PlanInterface
     /**
      * Returns whether it's possible to switch to this plan from a different plan.
      */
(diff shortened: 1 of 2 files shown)


AugustMiller:
@timkelty noted that renaming arguments of non-private methods _is_ actually a breaking change for PHP 8.0+, due to [named arguments](https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments).

So, maybe this needs to target 5.0?

lukeholder:
Added todo into the code for next major release

dcb84ee37987fcfe179d7cafdb777b28578ed75b

PR #3235, opened 2023-07-28
[4.x] Additional Generators

AugustMiller:
I was exploring some stuff in the Commerce extension docs and figured it was worth implementing [Generators](https://craftcms.com/docs/4.x/extend/generator.html) for some of the common components. My primary motivation is to have some external examples to point at, and to move some of the fussy out-of-context method signature stuff into code that can be maintained with the package, rather than as external documentation.

This introduces three new options, when the Generator is installed:

- Adjusters (`craft make adjuster`): Stubs and registers a new Order Adjuster class in the specified module or plugin;
- Gateways (`craft make gateway`): Stubs and registers a new Gateway class in the specified module or plugin (as well as creating the appropriate payment form and request/response models, and the corresponding back- and front-end templates);
- Shipping Method (`craft make shipping-method`): Stubs a new shipping method in the specified module or plugin (developer is responsible for registering it, as the logic for this is significantly different based on whether you just need a single hard-coded method or many methods—say, defined by and loaded from a third-party service);

Submitting as a **Draft**, because this definitely needs a second set of eyes… but I'm also not confident this covers even a fraction of the extension surface area. Purchasables are probably the biggest omission—and I don't even know where to begin with Tax Engines! 😂

### Todo:
- [x] Adjusters
- [x] Gateways
- [x] Shipping Methods
- [ ] Purchasables (unsure of workload here—seems like it would be pretty similar to `make element-type`…)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,7 @@
     "codeception/module-yii2": "^1.0.0",
     "craftcms/ckeditor": "^3.0",
     "craftcms/ecs": "dev-main",
+    "craftcms/generator": "^1.5",
     "craftcms/phpstan": "dev-main",
     "craftcms/rector": "dev-main",
     "craftcms/redactor": "*",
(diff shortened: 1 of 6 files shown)


AugustMiller:
Ah, interesting—PHPStan fails because `craftcms/generator` probably needs to be in `require-dev`?

PR #3236, opened 2023-08-01
Example Template Input Value

AugustMiller:
Noticed some Tailwind classes spilling into an `input`’s value on the `shop/cart/load.twig` template.


AugustMiller:
Nope, branched from the wrong spot. 😩

PR #3264, opened 2023-09-05
Order element docblock

AugustMiller:
Via support—a developer noticed that `Order::availableShippingMethods` was [still in the class reference](https://docs.craftcms.com/commerce/api/v4/craft-commerce-elements-order.html#availableshippingmethods), despite the method being removed in 4.0.

This removes the offending property hint, and adjusts the new `availableShippingMethodOptions` getter to be `property-read`.


lukeholder:
Thanks

PR #3267, opened 2023-09-07
fix: qualify `Emails::_createEmailQuery` `orderBy`

### LINKED ISSUE #3263: [3.x/4.x]: Emails::_createEmailQuery (and others) could be explicit about the orderBy
### What happened?

A slightly curly one, but in my test suite (running on SQLite in memory, you don't see this on MySQL) calling `\craft\commerce\services\Emails::getAllEmailsByOrderStatusId` fails with the following:

```
yii\db\Exception: SQLSTATE[HY000]: General error: 1 ambiguous column name: name
Failed to prepare SQL: SELECT `emails`.`id`, `emails`.`name`, `emails`.`subject`, `emails`.`recipientType`, `emails`.`to`, `emails`.`bcc`, `emails`.`cc`, `emails`.`replyTo`, `emails`.`enabled`, `emails`.`templatePath`, `emails`.`plainTextTemplatePath`, `emails`.`uid`, `emails`.`pdfId`, `emails`.`language` FROM `commerce_emails` `emails` INNER JOIN `commerce_orderstatus_emails` `statusEmails` ON `emails`.`id` = `statusEmails`.`emailId` INNER JOIN `commerce_orderstatuses` `orderStatuses` ON `statusEmails`.`orderStatusId` = `orderStatuses`.`id` WHERE `orderStatuses`.`id`=:qp0 ORDER BY `name`
```

This boils down to calling:

```
->orderBy('name')
```

without specifying the table (like so):

```
->orderBy('emails.name')
```

at:

https://github.com/craftcms/commerce/blob/3.4.22.1/src/services/Emails.php#L913
https://github.com/craftcms/commerce/blob/4.2.11/src/services/Emails.php#L894

when used in scenarios where a join against a table which also has a `name` column (such as `commerce_orderstatuses` in this case).

It seems like it would be a good idea (and help out my specific case :D) to provide the table alias to the `orderBy` call here (and perhaps anywhere else like this).

### Craft CMS version

3.8.9

### Craft Commerce version

3.4.21

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

msbit:
Add the table alias to the `orderBy` to avoid driver specific ambiguity in the generated SQL when joining against another table that _also_ has a `name` column.

This is in line with `States::_createStatesQuery`, and should fix #3263.

### DIFF
--- a/src/services/Emails.php
+++ b/src/services/Emails.php
@@ -891,7 +891,7 @@ class Emails extends Component
                 'emails.to',
                 'emails.uid',
             ])
-            ->orderBy('name')
+            ->orderBy('emails.name')
             ->from([Table::EMAILS . ' emails']);
     }
 


lukeholder:
@msbit thanks this is now in 4.3.1 out now.

PR #3273, opened 2023-09-12
Relocate external payment source links for Stripe

AugustMiller:
These links were just kind of floating out in the markup. Cleaned up the template a bit and made the offsite links read a bit better in more contexts.


lukeholder:
thanks

PR #3274, opened 2023-09-12
Optionally allow a shippingMethod to be associated with a ShippingMethodOption

### LINKED ISSUE #3271: [4.x]: Shipping Method Options no longer allow extra properties/data
### What happened?

Referencing [this](https://github.com/craftcms/commerce/issues/1378) issue and the [Postie](https://github.com/verbb/postie/issues/112) plugin.

Postie registers shipping methods, that are attached with `ShippingMethods::EVENT_REGISTER_AVAILABLE_SHIPPING_METHODS`. We use our own `ShippingMethod` class that extends the Commerce `ShippingMethod` class. In this, we store information about the carrier used (like UPS), and for each rate and service returned from the API, we store additional information like estimated delivery - not just the price.

What Commerce does is grab all available shipping methods, and create `ShippingMethodOption` models. I'm not sure _why_ this is done, but it's been this way since Commerce 3.x I believe. Users then use `cart.availableShippingMethodOptions` on their shipping page.

The issue we have is that through this translation of `ShippingMethod` to `ShippingMethodOption`, we lose Postie's extra information. Despite `ShippingMethodOption` also extending from a `ShippingMethod` class, it's the _Commerce_ class that's extended, and not Postie's. As such, any reference to the extra things Postie does is lost.

Might I suggest recording the original `ShippingMethod` that was used to create the `ShippingMethodOption` be added? We could add the following [here](https://github.com/craftcms/commerce/blob/41d24b82560acfb45c24ddbd66a75cf8e98a45ea/src/elements/Order.php#L2004):

```php
$option->shippingMethod = $method;
```

Which would store Postie's shipping method against the `ShippingMethodOption` for people to access in their templates.

Other options are:
- Adding in an extra property like `data` for storing arbitrary data for a `ShippingMethodOption`
- Adding an event for creating a snapshot - similar to Products/Variants and line items

### Craft CMS version

4.5.3

### Craft Commerce version

4.2.11

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

johnnynotsolucky:
### Description

Craft Commerce's `ShippingMethodOption` erases the information of any custom `ShippingMethod`s that were used to create them. While not ideal, this change sets a new field, `shippingMethod` on the `ShippingMethodOption` class and also proxies the call to `getShippingRules` to use the attached `ShippingMethod` if it is present.

Of course, other functions such as `getName`, `getType`, etc will not be overridden and would need to be accessed like `$option->shippingMethod->getType()`.

With this change the following snippet works with Commerce built-ins and custom shipping methods:

```twig
{% set cart = craft.orders().one() %}

{% for handle, method in cart.availableShippingMethodOptions %}
  <div class="shipping-select">
    <label>
      <input type="radio" name="shippingMethodHandle" value="{{ handle }}" {% if handle == cart.shippingMethodHandle %}checked{% endif %} />
      <strong>{{ method.name }}</strong>

      <span class="price">
          {{ method.getPrice() | commerceCurrency(cart.currency) }}
      </span>

      {% set shippingRules = method.getShippingRules() %}
      <code>{{ dump(shippingRules) }}</code>
    </label>
  </div>
{% endfor %}

```

### Related issues

- Resolves #3271

### DIFF
--- a/src/elements/Order.php
+++ b/src/elements/Order.php
@@ -1984,6 +1984,7 @@ class Order extends Element
             $option->handle = $method->getHandle();
             $option->matchesOrder = ArrayHelper::isIn($method->getHandle(), $matchingMethodHandles);
             $option->price = $method->getPriceForOrder($this);
+            $option->shippingMethod = $method;
 
             // Add all methods if completed, and only the matching methods when it is not completed.
             if ($this->isCompleted || $option->matchesOrder) {
--- a/src/models/ShippingMethodOption.php
+++ b/src/models/ShippingMethodOption.php
@@ -7,6 +7,7 @@
 
 namespace craft\commerce\models;
 
+use craft\commerce\base\ShippingMethod as BaseShippingMethod;
 use craft\commerce\behaviors\CurrencyAttributeBehavior;
 use craft\commerce\elements\Order;
 use craft\commerce\Plugin;
@@ -38,6 +39,20 @@ class ShippingMethodOption extends ShippingMethod
      * @var boolean
      */
     public bool $matchesOrder;
+
+    /**
+     * @var BaseShippingMethod|null The shipping method this option was derived from.
+     */
+    public ?BaseShippingMethod $shippingMethod = null;
+
+    public function getShippingRules(): array
+    {
+        if ($this->shippingMethod !== null) {
+            return $this->shippingMethod->getShippingRules();
+        }
+
+        return parent::getShippingRules();
+    }
 
     /**
      * @throws InvalidConfigException


lukeholder:
Fixed in 531a4d3c8019a87657b553d35273549d29acd7c2

```json
"require": {
  "craftcms/commerce": "dev-develop#531a4d3c8019a87657b553d35273549d29acd7c2 as 4.3.0",
  "...": "..."
}
```

Then run `composer update`.

Thanks!

lukeholder:
Decided to let you do `option.shippingMethod.getShippingRules()` since options are not shipping methods and not all shipping methods have rules. We plan to clean up this interface in the future.

nfourtythree:

PR #3279, opened 2023-09-18
Add order context to isPurchasableShippable in class ShippingMethod

boboldehampsink:


lukeholder:
I also added the optional param to other places that have the context and call the method. Will be in the next release. Thanks.

nfourtythree:

boboldehampsink:
Thanks @nfourtythree, please let me know when it is in 3.x too, thanks!

nfourtythree:
This is in `3.4.23` which has just been released!

PR #3297, opened 2023-10-12
Update variant SKU and title before save instead of before validation

johnnynotsolucky:
### Description

A variants title and SKU are generated automatically based on their respective templates. For this to work correctly when writing code against Commerce APIs and not using validation, we need to ensure that those two side-effects are moved out of validation otherwise they will never be updated correctly without a CLI or CP invoked resave of a product.

This PR removes the `updateTitle`and `updateSku` functions to the `beforeSave` function so that validation of a variant isn't required to generate those two fields data.

### DIFF
--- a/src/elements/Variant.php
+++ b/src/elements/Variant.php
@@ -1127,11 +1127,6 @@ class Variant extends Purchasable
      */
     public function beforeValidate(): bool
     {
-        $product = $this->getProduct();
-
-        $this->updateTitle($product);
-        $this->updateSku($product);
-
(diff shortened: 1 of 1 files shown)


lukeholder:
Thanks for reporting this.

Instead of doing this only before saving, we have decided to do it in addition to before validation. See 75f4a0a8557aab3ae2cf5abd810a093ff65ee152. 

This means we ensure that it is generated in both cases.

This doesn't break anything since the SKU will not be generated again if it already exists.

nfourtythree:

PR #3325, opened 2023-11-07
Use subquery in hasVariant & hasProduct criteria to improve performance

boboldehampsink:
### Description

This change makes the `hasVariant` criteria apply a subquery instead of fetching and applying via PHP


lukeholder:
Thanks Bob, I am just checking with the team there was not a reason we didn't do this originally. Will merge once I confirm.

lukeholder:
Thanks Bob!

PR #3327, opened 2023-11-08
Fix discounts that require a user to be logged in

### LINKED ISSUE #3326: [4.x]: Limited Use Coupon Code Works With Guest Checkout
### Description

Per User Discount Limit is not honored on guest checkout.

>Limit how many times, per user, the discount is allowed to be applied. This requires a user to be logged in; guests users will not be able to use the discount.

Doesn't appeart o be true at the moment.

### Steps to reproduce

1. Check out as a guest and enter a coupon code where Per User Discount Limit is set to 1

### Expected behavior

Shouldn't be applied if Per User Discount Limit is set to 1

### Actual behavior

Discount is applied.

### Craft CMS version

Pro 4.5.9

### Craft Commerce version

4.3.1

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

lukeholder:
### Description

Potentially fixes #3326


lukeholder:
Looking into fixing tests...

PR #3330, opened 2023-11-09
Bugfix/13911 order edit duplicate tabs

i-just:
### Description
On the order edit page, we’re getting both dynamic and static versions of the edit form. Each version pulls in the content tabs, so they’re essentially duplicated.

Issue 1: both static and dynamic tabs were taken into account when calculating if the tabs dropdown should be shown.
Fixed by re-initialising tabs after `OrderActions` had a chance to add a `hidden` class to the ones that should be hidden.

Issue 2: all tabs (for the static and dynamic forms) were showing in the tabs dropdown (looks like this existed in v3, too). Fixed by programmatically showing/hiding dropdown tab links depending on whether the order is being edited or not.

### Related issues
https://github.com/craftcms/cms/issues/13911


lukeholder:
Thanks @i-just !

lukeholder:
I will let @nfourtythree review/merge this as he has recent experience with the order editor vue app.

nfourtythree:
Thanks for this @i-just !

PR #3378, opened 2024-01-29
Fix deprecated typo in Order.php

joshuapease:
### Description

Minor fix, but lets PHPStorm and code linters find this deprecated method.


lukeholder:
Thanks!

PR #3380, opened 2024-02-05
Fix v4 upgrade psql issues

gijsstegehuis:
### Description
Found these SQL issues while running the commerce upgrade command on PostgreSQL 13.2


angrybrad:
Thank you!

PR #3442, opened 2024-03-26
[4.x] Support displaying json in transactions note field

samhibberd:
### Description

We use the transactions model note property to store a json string, been super helpful providing extra detail to our app where we need to extend the refund logic.

Issue is the cp tooltip can't parse that json note, this pr adds a little tweak to handle this and means our cp users can see the info they need:

Screenshot 2024-03-26 at 10 38 20

Screenshot 2024-03-26 at 09 28 13

### DIFF
--- a/src/controllers/OrdersController.php
+++ b/src/controllers/OrdersController.php
@@ -1518,6 +1518,15 @@ class OrdersController extends Controller
                     $transactionResponse = Json::htmlEncode($transactionResponse);
                 }
 
+                $transactionNoteType = 'text';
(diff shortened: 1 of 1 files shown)


samhibberd:
Is this something that might get considered for merge?

lukeholder:
@samhibberd we had a discussion on this and decided that we would prefer to make a data notes column that is specifically for use for json data. We will close this PR but have added the item to the backlog for the next release.

Thanks

samhibberd:
Sounds good 👌

lukeholder:
Thanks for your understanding. 🙏

PR #3478, opened 2024-05-02
Corrected translation for Number (Order Number)

NathanReul:
### Description
It is currently translated to "Aantal", which translates to "Amount" in english. In this context, the correct translation should be "Nummer"


lukeholder:
Thanks!

PR #3487, opened 2024-05-07
Fix Commerce 5 for Postgres

### LINKED ISSUE #3486: [5.x]: Can't install Commerce 5 with PostgreSQL
### Description

Can't install Commerce 5 with PostgreSQL, probably because the "siteId" column isn't properly quoted (as Postgres is case-sensitive):

```
web-1  | *** applying m230110_052712_site_stores
web-1  |     > create table {{%commerce_site_stores}} ...Exception: SQLSTATE[42703]: Undefined column: 7 ERROR:  column "siteid" named in key does not exist at character 211
web-1  | The SQL being executed was: CREATE TABLE "commerce_site_stores" (
web-1  |        "siteId" integer,
web-1  |        "storeId" integer NULL DEFAULT NULL,
web-1  |        "dateCreated" timestamp(0) NOT NULL,
web-1  |        "dateUpdated" timestamp(0) NOT NULL,
web-1  |        "uid" char(36) NOT NULL DEFAULT '0',
web-1  |        PRIMARY KEY(siteId)
web-1  | ) (/app/user/vendor/yiisoft/yii2/db/Schema.php:676)
```

### Steps to reproduce

Install Commerce 5 on PostgreSQL

### Craft CMS version

5

### Craft Commerce version

5

### PHP version

8.3

_No response_

### Database type and version

PostgreSQL 15.3

### Image driver and version

_No response_

-

boboldehampsink:
### Description

Properly escapes column names so it works on PostgreSQL as well (which is case-sensitive)

### Related issues

Closes #3486

### DIFF
--- a/src/migrations/m221122_055725_multi_store.php
+++ b/src/migrations/m221122_055725_multi_store.php
@@ -32,7 +32,7 @@ class m221122_055725_multi_store extends Migration
                 'dateCreated' => $this->dateTime()->notNull(),
                 'dateUpdated' => $this->dateTime()->notNull(),
                 'uid' => $this->uid(),
-                'PRIMARY KEY(id)',
+                'PRIMARY KEY[[(id]])',
             ]);
 
             $this->addForeignKey(null, Table::STORESETTINGS, ['id'], Table::STORES, ['id'], 'CASCADE', 'CASCADE');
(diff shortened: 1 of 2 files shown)


lukeholder:
Thanks, will get to your other issue shortly.

PR #3494, opened 2024-05-10
Fix duplicate column error in migration

alexmglover:
### Related issues

Error when running migration on 5.0.5

```Column already exists: 1060 Duplicate column name 'autoSetPaymentSource'```


lukeholder:
This was fixed in cf03b5111e672d0fdad83c6365ea8513e99d97fa and will be in the next release. Thanks!

PR #3500, opened 2024-05-15
Update Dummy.php expiry date in 4.x

MatMc:
### Description

Update the expiration date of the Dummy payment gateway's dummy credit card to always be in the next year, matching how the latest version of Commerce does it. Currently it requires manually changing the year on every use due to being in the past.


nfourtythree:
Hi @MatMc

Thank you for this!

PR #3507, opened 2024-05-21
disabled variant block should still be clickable

i-just:
### Description
If you’re on CMS 4.9.3, if you disabled a variant block, you couldn’t click on the cog to enable it. You couldn’t double-click the title bar to collapse/expand it, and you couldn’t select it via a checkbox.

Fixed by applying the same solution as used on cms v5 (https://github.com/craftcms/cms/pull/13897).
A [similar PR](https://github.com/craftcms/cms/pull/15004) was raised and merged for cms v4.

### Related issues
https://github.com/craftcms/cms/issues/15002


nfourtythree:
Thanks @i-just !

PR #3549, opened 2024-06-20
Fix Attempt to read property "id" on array error

joostwaaijer:
### Description

Fixes an error when trying to set estimatedBillingAddress with an array (added the same line that's included in setEstimatedShippingAddress)


lukeholder:
Thanks!!

boboldehampsink:
@lukeholder any idea when this is getting released?

lukeholder:
@boboldehampsink in about 5 hours.

PR #3550, opened 2024-06-21
[4.7] Add warning alert if upgrade command is not completed

lukeholder:
### Description

Need to get a better warning message before merging this:

> Commerce 4 Upgrade Incomplete. Please ensure the Commerce 4 upgrade command has finished running. Contact a developer if you are unsure.


src/Plugin.php
```diff
@@ -270,12 +271,34 @@ public function init(): void
...
+            if (Craft::$app->getDb()->columnExists($column['table'], $column['column'])) {
+                $columnsExist = true;
+                break;
+            }
+        }
+
+        if ($columnsExist) {
+            Event::on(Cp::class, Cp::EVENT_REGISTER_ALERTS, static function($event) {
+                $event->alerts[] = [
+                    'content' => Craft::t('commerce', '<strong>Commerce 4 upgrade incomplete.</strong> Please ensure the <a href="https://craftcms.com/docs/commerce/4.x/upgrading.html#performing-the-upgrade" target="_blank" rel="noopener noreferrer">Commerce 4 upgrade command</a> has finished running.'),
```
gcamacho079: @lukeholder just popping in because I recently started updating some translations that include brand names and saw that this was relevant. Because the brand names aren’t translated, I’ve started wrapping them with `<span lang="en"></span>` so that screen readers switch their pronunciation to English when encountering them in the context of another language.

You can see an example of this in [this recent commit](https://github.com/craftcms/cms/pull/14935/commits/6c0675949af80a224ddf195f405fe99e7f397560), and Brandon also provided some [guidance here about ensuring they're not being escaped](https://github.com/craftcms/cms/pull/14935#discussion_r1589575022).

lukeholder: thanks @gcamacho079 ! will do.

PR #3566, opened 2024-07-03
[4.x] Shipping method performance issue due to order serialization.

lukeholder:
### Description

Discovered this is the performance issue from blackfire profiling.


lukeholder:
Please see #3653

PR #3588, opened 2024-07-15
Fix migration for Postgres

timkelty:
### Description

Upgrading from commerce 4 to 5 on Postgres 16:

```
*** applying m231110_081143_inventory_movement_table
Exception: SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: boolean = integer
LINE 3: WHERE "primary"=1
                       ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
The SQL being executed was: SELECT "id"
FROM "commerce_stores"
WHERE "primary"=1
```


lukeholder:
Changelog item: ee5dd91caf7f78094e0c00d83b23d1e6dc66b41d

thanks!

lukeholder:
This was released in Craft Commerce 5.0.13  Thanks!

PR #3601, opened 2024-07-24
Fixes another piece of this migration for Postgres

kennethormandy:
### Description

Hey! Best I can tell, there was one more spot that `['primary' => 1]` needed to change to `['primary' => true]` for Postgres in the migration in #3588.

I ran into this independently, but this resolves the issue for me.


lukeholder:
Thanks!

PR #3604, opened 2024-07-26
Fixes custom Stripe button text param in example templates

kennethormandy:
### Description

Hey! Best I was able to tell, you need to customize the Stripe button text using `submitButtonText`, not `submitButtonLabel`.

[It’s correct](https://github.com/craftcms/commerce-stripe?tab=readme-ov-file#submitbuttonclasses-and-submitbuttontext) in the Stripe gateway README, but not in these example templates. The default is already “Pay” so it’s not obvious until you try and change it.


lukeholder:
Thanks!

PR #3619, opened 2024-08-08
Allow non-Craft queues

timkelty:
### Description

If you use a non-Craft queue component, e.g. `yii\queue\sqs\Queue`, you'll get an excption like this:

```
craft\commerce\services\CatalogPricing::generateCatalogPrices(): Argument #4 ($queue) must be of type craft\queue\Queue|craft\queue\QueueInterface|null, yii\queue\sqs\Queue given, called in /var/app/current/vendor/craftcms/commerce/src/queue/jobs/CatalogPricing.php on line 37
```


nfourtythree (approved):

nfourtythree:
Commerce `5.0.16` has now been released which includes this fix.

Thanks!

PR #3620, opened 2024-08-08
Fixed widgets based on thisWeek date range

vnali:
### Description
Hi.
I noticed that changing Week Start Day on user preferences does not apply to widgets based on thisWeek and a week always start from Monday on widgets because `$this->weekStartDay` is set after `$this->_setDates();`.
This fix moves `$this->weekStartDay` to top of `__construct` function. 
I tested this and it seems it works for `Total Orders` widget and its charts start correctly from the day that user is selected as week start day.


nfourtythree (approved):
Good catch, thanks for this @vnali !

nfourtythree:
Hi

Commerce `5.0.16` has now been released which includes this fix.

Thanks!

PR #3628, opened 2024-08-15
[5.x] Reports

lukeholder:
### Description

- [ ] Reports Section
- [ ] Reports base class


lukeholder:
Reports will be in core Craft CMS in the future.

PR #3633, opened 2024-08-15
Don't log uncategorized exceptions

timkelty:
### Description
You might argue these logs shouldn't be there at all, since the exceptions are already thrown – but short of that, this at least gives the logs a category so they can be filtered out.

Without this, you'll get error logs for expected things like "Insufficient funds" with no good way to filter them.


nfourtythree:
Need to put these changes in `4.x` so have merged them in from there in this commit: d6de5339eb5afea18a9f8059c8b9528d0f727bac

Issue #3132, opened 2023-04-05
[4.x]: Discount automatically applied to all orders when no coupon code is set


samueldraper:
### Description

Users automatically seeing coupon codes applied at checkout, and discount applied for which they are not eligible.

Cause seems to be because no coupon code was specified therefore (one of several but not all) discounts were applied to cart.

### Steps to reproduce

1. Create discount rules
2. Do not add a coupon code

### Expected behavior

Should not apply to all carts.
Coupon code should be mandatory.

### Other info

Based on  #3131 this issue **is actually** expected behaviour. If that is the case a UX enhancement would be much needed in order to avoid the possibility of this happening. Even as simple as some validation: "Are you sure you don't want to add a coupon code? This will result in discount being applied to all carts that meet the criteria".

### Craft CMS version

4

### Craft Commerce version

4

### Installed plugins and versions

-

lukeholder:
Coupon codes are just a condition for the discount, just like any other condition on the Discount. 
The feature is `Promotions > Discounts`, not `Promotions > Coupons`.

We could make it cleaner within the discount edit page, though. This is the first time someone has submitted this as a misunderstanding. I will leave the ticket open and see what I can do.

soulseekah:
We've lost almost 1000 Euros due to this misunderstanding. A content manager deleted the coupon code but not the discount itself...

This "feature" is what Sales are supposed to do, it's highly counterintuitive and should probably be behavior that's off by default (on for existing Discounts).

+1 on at least making it clearer on the page that the discount is automatically applied.

The documentation does hint at this behavior, but does not state this specifically as potentially unwanted behavior.

P.S.: Such discounts actually stack.

Romanavr:
@lukeholder 

This is completely unexpected behaviour, and I see no reason why the coupon code should be automatically applied when no coupon code is generated? Could you give me some real practical examples where it should be used?
I accidentally ran into the same issue and I still confused. We lost some money because of it.

We have sales for the _automatic_ discounts, and this is logically they can be applied to any customer. They are even located in a separate section. But I'll be honest, I don't understand at all why a coupon code works as a sale, especially when there is no actual coupon code lol.

Could you please reconsider this behaviour of coupon codes and ,maybe, mark it as a bug? Or perhaps I don't know the reason why it's done in such way.

Either way, thanks for any answer!

nfourtythree:
Hi all

[...]

PR #3660, opened 2024-09-06
[4.7] Add require coupon code lightswitch to discounts

nfourtythree:
### Related issues
#3132


nfourtythree:
@AugustMiller would love to get your input on the wording of "Require Coupon Code". Toyed with the idea of "Requires" but then felt like it should be "Requires a Coupon Code" which seems lengthier. 

The idea behind the PR being this is now an extra fail safe that you explicitly say whether the discount needs a coupon code to apply rather than simply an empty coupons table. Which causes issue if someone was to accidentally delete all the codes.

CleanShot 2024-09-10 at 08 04 22

lukeholder (approved):
LGTM

PR #3663, opened 2024-09-09
[5.2] Orderable product types

lukeholder:
### Description

This PR enabled product types to have 2 types:

- Channel
- Orderable

Channel is the default, and orderable works like entry structures with a locked max depth of 1 (not nestable like structures).


BrandKits:
This sounds great. Do you see any issues changing existing product types from Channel to Orderable? And vice versa.

Would love a similar ability to solve ‘Add ability to reorder Line Items when creating/editing CP Orders (Discussion #3597)’.

sjcallender:
I've tested this and it works well. Stoked to have this finally. 🚀

BrandKits:
A key use case of this is to order products globally, but also by collection (entry). The current product field workaround is proving very hard. With hundreds of products per entry adding and dragging new products from bottom to top is painful. Hopefully like structures you can ‘add products above’.

It would be very valuable if there was a reverse relationship field that can show all entry products that are related, AND be able to order them manually beyond the global sort order.

Is there a beta that we can test or do we need to wait for 5.2? Is that expected soon?

Thanks

Issue #3306, opened 2023-10-19
[4.x]: New address in order entry should only show enabled countries


bossanova808:
### Description

Not a huge or urgent thing, but would be a good time saver - I feel the 'New Address' entry in order processing should only show the enabled countries.

I.e. I think this is inconsistent:

![image](https://github.com/craftcms/commerce/assets/731309/ef18561d-5767-4bf3-ac77-52b2f6bb9675)

But then it defaults to (of course) the US.  I feel it should default (in our case) to the store country (Australia) and only list enabled countries:

![image](https://github.com/craftcms/commerce/assets/731309/af2200e8-33fe-4571-814a-b7f571fd7179)

### Craft CMS version

Craft Pro 4.5.7

### Craft Commerce version

4.3.1

### PHP version

8.2

### Operating system and version

NA

### Database type and version

NA

### Image driver and version

NA

### Installed plugins and versions

NA

nfourtythree:
Hi @bossanova808 

Thank you for your great suggestion, this is something we will definitely take on board and look to get in, in the future.

There are a few things to think about with this, for example, there is a business case that would only want to allow certain countries in the front end for people to check out. But maybe over the phone or internally store administrators may want to create orders for countries outside of that list. We also need to take into consideration what data Craft allows Commerce to pass to the instantiation of the new address element.

Definitely something we are looking to improve on and we will keep this issue updated with any/all developments.

Thanks!

bossanova808:
Good points - how about a compromise solution in the interim - at least _default_ the new address country to the store country, and list those others enables at the _top_ of the list, with the rest then underneath?

This would allow the most common cases to be much quicker, but still give the flexibility to do the things you mention...

bossanova808:
Following up here - any chance you can do this?  It really is very irritating and surely jsut making the store's country the default selected for new addresses is not controversial?

nfourtythree:
Commerce `4.7.0` and `5.2.0` have now been released. When editing an order and creating a new address the default country selected will now be the same country as the store's location.

The upcoming Craft `4.13` will include an event which will allow you to restrict the countries available for selection when creating an address.

Thanks!

brandonkelly:
Craft 4.13.0 and 5.5.0 are out now with that new event (https://github.com/craftcms/cms/pull/15711).

PR #3665, opened 2024-09-10
[4.7] Default selected country is the same as the store's location when creating a new address while editing an order

nfourtythree:
### Description
With this PR when creating a new address for an order on the order edit page, the country that is initially selected on the address form is the same country as the store's location address.

### Related issues
#3306


lukeholder (approved):
LGTM

PR #3675, opened 2024-09-16
[4.7] Upgrade event

lukeholder:
### Description

Added the ability to hook into the upgrade command after it has run, but before v3 tables and columns are dropped.

Event Model: `\craft\commerce\events\UpgradeEvent`
Event: `\craft\commerce\console\controllers\UpgradeController::EVENT_BEFORE_DROP_V3_DATABASE_ENTITIES`

Event includes the column mapping so previous address IDs can be accessed and migrated by custom module or plugin code: https://github.com/craftcms/commerce/blob/6dfad2bb6150064bae2247f8bd74e16b7e79de50/src/console/controllers/UpgradeController.php#L99-L128


nfourtythree (approved):
Checked it, tested I could access the event etc. Looks good to me!

Issue #3689, opened 2024-09-24
Variant search condition rule


nfourtythree:
### Description

To enhance flexibility in filtering the product index, introduce a new condition rule that features a plain text field for searching variants. 

This input will be utilized in a variant query, enabling more precise product limitations based on the search results.

### Craft CMS version

4.x/5.x

### Craft Commerce version

4.x/5.x

nfourtythree:
Commerce versions `4.7.0` and `5.2.0` have now been released with this feature included.

PR #3691, opened 2024-09-24
[4.7] Add "Variant Search" condition rule to product conditions 

nfourtythree:
### Description
To enhance flexibility in filtering the product index, introduce a new condition rule that features a plain text field for searching variants.

This input will be utilised in a variant query, enabling more precise product limitations based on the search results.

### Related issues
#3689


nfourtythree:
Example of condition rule in action:


https://github.com/user-attachments/assets/f42c8644-1ce1-4654-a701-8e3ce1e8d8ca

Issue #3769, opened 2024-11-15
[5.x]: Reordering of disabled variants


i-just:
### What happened?

> We also have a different problem when I move an disabled variant:
> ![Image](https://github.com/user-attachments/assets/90cc285a-d517-40e0-933b-915512feabfc)
> And save the page with "ctrl" + "s" the whole layout of variants changes:
> ![Image](https://github.com/user-attachments/assets/1536fd61-aa9a-4396-a545-b1429f2b85fc)
> In the first image, we initially had two disabled products stacked above each other, with our moved variant positioned below them. However, after saving the page, the arrangement changed. Now, either:
> - Two enabled variants have moved up, or
> - Our moved variant has shifted down, now paired with another disabled variant.
> 
>  

 _Originally posted by @WHITE-developer in [#16103](https://github.com/craftcms/cms/issues/16103#issuecomment-2476010615)_

### Craft CMS version

5.5.0.1

### Craft Commerce version

5.2.4

### PHP version

8.3

### Operating system and version

ddev

### Database type and version

mysql 8.0

### Installed plugins and versions

-

WHITE-developer:
### Some additional info:

Occasionally, variants cannot be moved up to a new position down works fine but up sometimes won't work. In the attached video, you can see the following behavior:

The first variant I moved up successfully retained its new position after saving.
However, when I attempted to move another variant up, it reverted to its original position after saving.

### Steps to Reproduce:

1. Open the variants.
2. Move one variant to a new position and save.
3. Move another variant to a different position and save.
4. Observe whether the second variant retains its position.

### Expected Behavior:
- All moved variants should retain their new positions after saving.

### Actual Behavior:

- While some variants stay in their new positions, others revert to their original positions after saving.

### Attachments:
When there is nothing happening on screen I'm saving the page

https://github.com/user-attachments/assets/7710c724-2eed-4b60-8237-e7a0917efbcb

lukeholder:
@WHITE-developer we have a potential fix for this on this PR: https://github.com/craftcms/commerce/pull/3779

We are just testing it a little more before including it in a release.

WHITE-developer:
Hi @lukeholder,

Are there’s any updates regarding this release? Our client is currently awaiting a fix, and I’d appreciate any information or timeline you could provide.

lukeholder:
@WHITE-developer this was just fixed in the release last night. Craft Commerce 5.2.8

PR #3772, opened 2024-11-15
when getting variants for NestedElementsController, include disabled too

i-just:
### Description
When the `nested-elements/reorder` action is called for product variants, the `beforeAction()` method calls `getVariants()` via [magic getter](https://github.com/craftcms/cms/blob/5.5.0.1/src/controllers/NestedElementsController.php#L69) and a collection of elements is returned. By default, `getVariants()` doesn’t grab disabled elements. 

For reordering to work in that case, the disabled elements have to be returned, too.

(I opted to amend the getter for the other NestedElementsController action, too (delete). Commerce doesn’t seem to use it at the moment, but just in case.)

### Related issues
#3769

### DIFF
--- a/src/elements/Product.php
+++ b/src/elements/Product.php
@@ -105,6 +105,22 @@ class Product extends Element implements HasStoreInterface
     /**
      * @inheritdoc
      */
+    public function __get($name)
+    {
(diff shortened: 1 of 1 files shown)


lukeholder:
Fixed by #3779

Issue #3785, opened 2024-11-26
[5.x]: "Max Levels" are not taken into account for structured product types


HigumaSan4050:
### Description
Enabled lightswitch for "Enable structure for products of this type" and "Max Levels" set to 1 has no effect on the product lists.You can still move products under other products

### Steps to reproduce

1. create a "Product Type" with "Enable structure for products of this type" enabled and "Max Levels" set to 1
2. Create 2 Products in this Product Type
3. You can move one of the products unter the other product

### Craft CMS version

5.5.3

### Craft Commerce version

5.2.5

### PHP version

8.3

nfourtythree:
Hi @HigumaSan4050 

Thank you for reporting this, we were able to reproduce the issue and are working on a fix.

Thanks

nfourtythree:
Hi @HigumaSan4050 

We have pushed a fix for this issue which will be included in the next release of Commerce

To get this early, change your `craftcms/commerce` requirement in your project's `composer.json` to:

```json
"require": {
  "craftcms/commerce": "5.x-dev as 5.2.6",
  "...": "..."
}
```

Then run `composer update`.

Thanks!

PR #3787, opened 2024-11-28
[5.x] Product type’s max levels setting not being respected

nfourtythree:
### Related issues
#3785

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -257,7 +257,7 @@ class Plugin extends BasePlugin
     /**
      * @inheritDoc
      */
-    public string $schemaVersion = '5.2.0.6';
+    public string $schemaVersion = '5.2.7.0';
 
     /**
      * @inheritdoc
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -257,7 +257,7 @@ class Plugin extends BasePlugin
     /**
      * @inheritDoc
      */
-    public string $schemaVersion = '5.2.0.5';
+    public string $schemaVersion = '5.2.0.6';
 
     /**
      * @inheritdoc
--- a/src/migrations/m241128_174712_fix_maxLevels_structured_productTypes.php
+++ b/src/migrations/m241128_174712_fix_maxLevels_structured_productTypes.php
@@ -2,7 +2,6 @@
 
 namespace craft\commerce\migrations;
 
-use Craft;
 use craft\db\Migration;
 use craft\db\Query;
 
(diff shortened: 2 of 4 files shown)


src/Plugin.php
```diff
@@ -257,7 +257,7 @@ public static function editions(): array
     /**
      * @inheritDoc
      */
-    public string $schemaVersion = '5.2.0.5';
```
brandonkelly, line 260: The first 3 numbers here should correspond with the plugin version it will be released as. So `5.2.7.0`.

src/migrations/m241128_174712_fix_maxLevels_structured_productTypes.php
```diff
@@ -0,0 +1,41 @@
...
+class m241128_174712_fix_maxLevels_structured_productTypes extends Migration
+{
+    /**
+     * @inheritdoc
+     */
+    public function safeUp(): bool
+    {
+        // Get all product types from Commerce project config
+        $structuredProductTypesWithMaxLevels = (new Query())
+            ->from('{{%commerce_producttypes}}')
```
brandonkelly: Should probably be `Table::PRODUCTTYPES`.

src/migrations/m241128_174712_fix_maxLevels_structured_productTypes.php
```diff
@@ -0,0 +1,41 @@
...
+            ->where(['isStructure' => 1])
+            ->andWhere(['not', ['maxLevels' => null]])
+            ->collect();
+
+        // Loop through and update the `maxLevels` column in the `structures` table
+        $structuredProductTypesWithMaxLevels->each(function($productType) {
+            $this->update('{{%structures}}', ['maxLevels' => $productType['maxLevels']], ['id' => $productType['structureId']]);
```
brandonkelly: Should be `\craft\db\Table::STRUCTURES`

src/migrations/m241128_174712_fix_maxLevels_structured_productTypes.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    /**
+     * @inheritdoc
+     */
+    public function safeUp(): bool
+    {
+        // Get all product types from Commerce project config
+        $structuredProductTypesWithMaxLevels = (new Query())
+            ->from('{{%commerce_producttypes}}')
+            ->where(['isStructure' => 1])
```
brandonkelly: `1` should be `true`. Otherwise this will break in Postgres.

PR #3809, opened 2024-12-10
[5.3] Products & Variants card view designer

nfourtythree:


nfourtythree:
Commerce 5.3 has now been released with this included.

Thanks!

PR #3841, opened 2025-01-08
Improve shipping methods performance

boboldehampsink:
### Description

Currently commerce already calculates shipping methods total price before checking if it is enabled or matches the order. We have a lot of disabled shipping methods and this causes performance issues. Looking at the code I see no reason to move this line after the enabled/match checks.


lukeholder:
Thanks!

Issue #3848, opened 2025-01-14
[4.x]: Address validation causes internal server error in checkout


jornwildenbeest:
### Description

When trying to set a address that has empty required fields you get a internal server error in the checkout.

The error I am receiving in our checkout is:

```
2025-01-14 11:49:57 [web.ERROR] [craft\errors\InvalidElementException] Element 1726666 could not be duplicated because it doesn't validate. {"trace":["#0 /home/forge/site/releases/78/vendor/craftcms/commerce/src/controllers/CartController.php(655): craft\\services\\Elements->duplicateElement()","#1 /home/forge/site/releases/78/vendor/craftcms/commerce/src/controllers/CartController.php(252): craft\\commerce\\controllers\\CartController->_setAddresses()","#2 [internal function]: craft\\commerce\\controllers\\CartController->actionUpdateCart()","#3 /home/forge/site/releases/78/vendor/yiisoft/yii2/base/InlineAction.php(57): call_user_func_array()","#4 /home/forge/site/releases/78/vendor/yiisoft/yii2/base/Controller.php(178): yii\\base\\InlineAction->runWithParams()","#5 /home/forge/site/releases/78/vendor/yiisoft/yii2/base/Module.php(552): yii\\base\\Controller->runAction()","#6 /home/forge/site/releases/78/vendor/craftcms/cms/src/web/Application.php(341): yii\\base\\Module->runAction()","#7 /home/forge/site/releases/78/vendor/craftcms/cms/src/web/Application.php(640): craft\\web\\Application->runAction()","#8 /home/forge/site/releases/78/vendor/craftcms/cms/src/web/Application.php(303): craft\\web\\Application->_processActionRequest()","#9 /home/forge/site/releases/78/vendor/yiisoft/yii2/base/Application.php(384): craft\\web\\Application->handleRequest()","#10 /home/forge/site/releases/78/web/index.php(12): yii\\base\\Application->run()","#11 {main}"],"memory":41228816,"exception":"[object] (craft\\errors\\InvalidElementException(code: 0): Element 1726666 could not be duplicated because it doesn't validate. at /home/forge/site/releases/78/vendor/craftcms/cms/src/services/Elements.php:1646)"} 
```

This is caused due invalid addresses that are missing a required field for example.
We have some shops where we've improved the address forms in a later stadium, or made fields required that weren't required before. This leads to older addresses that are still missing the newly added required fields. 

[...]

PR #3854, opened 2025-01-20
[4.x] Bubble up validation errors to the cart when adding an invalid customer address

nfourtythree:
### Description
Previously this would cause a PHP exception.

### Related issues
#3848


lukeholder (approved):
lgtm

PR #3859, opened 2025-01-22
Error saving OrderHistory, userId must be initialized, fixed #3858

### LINKED ISSUE #3858: [4.x]: Error saving OrderHistory, userId must be initialized
### Description
When updating an Order Status through the console we are currently facing the following exception: 
`Typed property craft\commerce\models\OrderHistory::$userId must not be accessed before initialization`

This is a result by a previous fix that only sets the userId when the request is not a console-request  (https://github.com/craftcms/commerce/commit/be857b5f91c905f624c1fdb39c15c9c8786d77cd)

Therefor when doing the following, we get the error mentioned above.

```php
$order = ...
$order->orderStatusId = $statusCompletedId;
Craft::$app->getElements()->saveElement($order);
```

### Craft CMS version

4.13.8

### Craft Commerce version

4.7.2

### PHP version

8.3

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

rvgool-oym:
### Description

When updating an Order Status through the console we are currently facing the following exception:
Typed property craft\commerce\models\OrderHistory::$userId must not be accessed before initialization

This fix sets the userId to null on initialization.

### Related issues
#3858

### DIFF
--- a/src/models/OrderHistory.php
+++ b/src/models/OrderHistory.php
@@ -56,7 +56,7 @@ class OrderHistory extends Model
     /**
      * @var int|null User ID
      */
-    public ?int $userId;
+    public ?int $userId = null;
 
     /**
      * @var string|null User name or email


lukeholder:
Thanks for this. We have made the same fix on the 4.x and 5.x braches. See this explainer: https://github.com/craftcms/commerce/issues/3858#issuecomment-2629248360

Thanks again!

PR #3864, opened 2025-01-28
Persist `makePrimaryShippingAddress` and `makePrimaryBillingAddress` on carts

lukeholder:
### Description

When choosing addresses from a customer's address book to use as the addresses on the cart, `billingAddressId` or `shippingAddressId` are submitted.

Optionally, you can also pass `makePrimaryShippingAddress` or `makePrimaryBillingAddress`, which allows the customer to make these submitted addresses the new primary addresses in their address book.

Alternatively, instead of submitting address IDs from the address book, the customer can submit a new address directly to the cart. This address would only exist on the order and would not be added to the address book, even after checkout. However, customers requested the ability to collect a new address and have it added to the address book *after* checkout. 

To address this, we introduced cart parameters called `saveBillingAddressOnOrderComplete` and `saveShippingAddressOnOrderComplete`. These parameters can be added to the cart so that, when the order is completed, the addresses on the order are copied to the customer's address book.

The issue this PR fixes is related to the behavior of copying these addresses to the address book. Now, when this copy occurs, we will check the `makePrimaryShippingAddress` or `makePrimaryBillingAddress` flags to determine whether the copied addresses should become the primary addresses in the address book. Previously, these flags only worked when selecting an existing address from the address book.

The `makePrimaryShippingAddress` or `makePrimaryBillingAddress` were previously never persisted to the order record but they are now in this PR. This means during checkout the developer can know the customers selection during checkout (since previously they only existed as options within a single cart update request).


nfourtythree:
Commerce 5.3 has now been released with this included.

Thanks!

Issue #3815, opened 2024-12-13
[5.x]: Restoring a soft-deleted Product does not restore it's Variants


roberttolton:
### Description

When you soft-delete a Product, it's Variants are also soft-deleted. However, when you restore the Product, it's Variants remain soft-deleted and the UI offers no option to restore these Variants, leaving them permanently soft-deleted.

### Steps to reproduce

1. Create a Product with a Variant
2. Soft-delete the Product
3. Restore the Product
4. Try and restore the Product's Variant

### Expected behavior

Variants should be automatically restored upon their owning Product being restored and/or the UI should allow for the separate restoration of desired/all Variants.

### Actual behavior

Variants are left permanently soft-deleted and cannot be restored.

Screenshot:
[Image](https://github.com/user-attachments/assets/03f9bc34-93cd-4974-a1f3-9051fcbf5dae)

### Craft CMS version

5.5.6.1

### Craft Commerce version

5.2.9

### PHP version

8.2

lukeholder:
@roberttolton sorry for delay on response.

I initially couldn't reproduce this issue, but can now in our latest release of Craft Commerce 5.3. I will look into a fix and get back to you as soon as possible.

Thanks for reporting it.

lukeholder:
Thanks for reporting this. I have PR that fixes this here #3880

We will update you once it is included in a release. Thanks!

nfourtythree:
Hi all

Commerce `5.3.2` has now been released with this fix included.

Thanks!

PR #3880, opened 2025-02-04
Restore variants when product is restored

lukeholder:
### Description

Fixes the restore of variants when their product is restored.

The old `deletedWithProduct` never worked. The column in the variants table never was populated. `deletedWithOwner` is the correct property.

I have deprecated the property and marked it to be removed in next break-point.

### Related issues

https://github.com/craftcms/commerce/issues/3815


nfourtythree (approved):
Looks good!

PR #3882, opened 2025-02-05
Stock cache should be updated on movements

lukeholder:
### Description

Fixed updating of stock cache as well as fixing transfers that contain deleted inventory items.

### DIFF
--- a/src/collections/InventoryMovementCollection.php
+++ b/src/collections/InventoryMovementCollection.php
@@ -25,6 +25,7 @@ class InventoryMovementCollection extends Collection
 {
     /**
      * @return array
+     * @since 5.3.2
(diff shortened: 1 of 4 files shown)


src/collections/InventoryMovementCollection.php
```diff
@@ -22,4 +23,13 @@
  */
 class InventoryMovementCollection extends Collection
 {
+    /**
+     * @return array
+     */
```
nfourtythree, line 29:
```suggestion
     * @since 5.3.2
     */
```

nfourtythree (changes requested):
One minor update, otherwise looks good to me!

Issue #3895, opened 2025-02-13
[5.x]: LineItems aren't updated with update cart action `cart/update-cart`


roguesherlock:
### Description

When updating a cart, the line item is expected to update based on the `id` provided in the payload. However, the update fails because the code appears to use the array index instead of the provided `id`, preventing the correct line item from being updated.

### Steps to reproduce

1. Send a request to update the cart with the following JSON payload:
```json
   {
       "lineItems": [{
           "id": 129,
           "qty": 10
       }]
   }
```

2. Check the cart to see if the line item with id 129 has been updated.

### Expected behavior
The cart should update the line item with the provided id (in this case, 129), so that the line item reflects the new quantity. In this case, the updated line item should show a quantity of 10.

### Actual behavior
The cart message says cart updated but the line item is not updated.

I dug through the code a bit I see that it attempts to use the array index (0) as the line item ID instead of using the `id` field (129) from the line item object, causing the wrong item to be fetched or no item to be found.
https://github.com/craftcms/commerce/blob/75832abf491ba0632d713ace8343cff4b47ae0a9/src/controllers/CartController.php#L245-L246

if I send the line Items as an object of key (id) /value (lineItem) pairs like the code expects it seems to work

```json
   {
      "lineItems": {
        "129": {
            "id": 129,
            "qty": 10
        }
      }
   }
```

However the documentation states
> Array of one or more of the cart's line items to update. Each must have an id key-value pair, and may include options, note, and qty key-value pairs.

I'm not sure if it's the expected behaviour and the documentation is incorrect (or I misunderstood it) or it's actually a bug.

### Craft CMS version

Craft Pro 5.6.4

### Craft Commerce version

5.3.1

### PHP version

8.3.14

### Operating system and version

Darwin 24.3.0

### Database type and version

MySQL 8.4.4

### Image driver and version

Imagick 3.7.0 (ImageMagick 7.1.1-41)

lukeholder:
@roguesherlock You are correct the request expects `key` to be the line item ID.

We don't currently do anything with `"id": 129,` when you pass it in the data object at the moment. It is ignored.

Documentation of this behavior is here:

https://craftcms.com/docs/commerce/5.x/development/cart.html#updating-line-items

If we decided to make a change it would need to be in a backwards-compatible way.

One idea is to look in the line items object data for an ID first and use that, and if it is missing, defer to the key.

This would need to be in the next minor/major release and not in a patch.

[...]

PR #3898, opened 2025-02-17
[5.4] Look for ID being passed in lineItems param

lukeholder:
### Description

Backward compatible way to allow for a line item ID to be passed though within the object instead of relying on an object key.

### Related issues

https://github.com/craftcms/commerce/issues/3895

### DIFF
--- a/src/controllers/CartController.php
+++ b/src/controllers/CartController.php
@@ -243,15 +243,15 @@ class CartController extends BaseFrontEndController
 
         // Update multiple line items in the cart
         if ($lineItems = $this->request->getParam('lineItems')) {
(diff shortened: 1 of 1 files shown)


lukeholder:
We will look to update the docs, but we think keeping things working the same way it currently does (requiring the ID of the line item to be the key of the line item data to be updated).

You can see in the form example that the key in the form data is the line item ID:

https://craftcms.com/docs/commerce/5.x/development/cart.html#updating-line-items

We will look to add better docs for updating the cart over JSON so confusion like https://github.com/craftcms/commerce/issues/3895 is avoided.

Thanks!

Issue #3283, opened 2023-09-28
[4.x]: Internal adjustsments cache not updated after saving new adjustments?


boboldehampsink:
### What happened?

In the method \craft\commerce\elements\Order::_saveAdjustments, previous adjustments can be deleted. But if you run `$this->getAdjustments()` again after that, the deleted adjustments are still there? I think _saveAdjustments should update the adjustments with `setAdjustments` after it has done its part.

### Craft CMS version

4

### Craft Commerce version

4

### Installed plugins and versions

-

lukeholder:
I do not think this is the case.

In memory (contained within `Order::_orderAdjustments` and retrieved in `_saveAdjustments()`) the orders have the adjustments will be saved - it does not contain adjustments that will be deleted. We are purely deleting the old adjustment records which are no longer on the order.

Let me know if I am missing something.

boboldehampsink:
@lukeholder ok let me tell you where this is coming from. We have a LOT of exceptions of "No line item exists with the ID “...”" - so I'm guessing there is some sort of race condition somewhere? Any thoughts? Here is a full stacktrace: https://robuust-6t.sentry.io/share/issue/d0e80dab2b3d4655bff6e3ceeadd866c/

lukeholder:
@boboldehampsink I can't tell much from the stack trace apart from the error being triggered due to the line item referencing an ID that has been deleted. 

Do you have any custom module code manipulating the order or line items?

boboldehampsink:
Hi @lukeholder, no there is no custom code manipulating the order or line items. Also, this happens on various projects with Commerce.

boboldehampsink:
@lukeholder still seeing this error alot over various projects. I'll try to make it reproducible

boboldehampsink:
Could this be some sort of race condition? Imagine a front-end which updates the cart via an ajax request (by clicking a button on a product cart). What would happen if you click many different buttons rapidly?

lukeholder:
@boboldehampsink could you try this branch as a potential fix for this?

https://github.com/craftcms/commerce/pull/3879

I think using the Active record to delete could come with a race condition, so we can just rely on the bulk delete (that is just below the code removed). Thanks!

boboldehampsink:
> [@boboldehampsink](https://github.com/boboldehampsink) could you try this branch as a potential fix for this?
> 
> [#3879](https://github.com/craftcms/commerce/pull/3879)
> 
[...]

PR #3901, opened 2025-02-20
FIxed race condition on deleting old adjustment records

lukeholder:
### Description

No need to delete the Record if we already drop the rows directly in the database.

### Related issues

https://github.com/craftcms/commerce/issues/3283


lukeholder:
To get this potential fix early and test out the PR, change your `craftcms/commerce` requirement in composer.json to:

```json
"require": {
  "craftcms/commerce": "dev-feature/pt-2407-fix-adjustment-id-race-condition as 4.8.1.2",
  "...": "..."
}
```

Then run `composer update`.

boboldehampsink:
@lukeholder any news on this?

boboldehampsink:
@lukeholder I want to test this on Craft 5, can you make an updated branch for that?

boboldehampsink:
Now seeing a lot of these errors so I can test it properly if you want

lukeholder:
@boboldehampsink Yes please, testing this would be useful. Updated to 5.x base branch.

boboldehampsink:
@lukeholder have been running this for a while but the errors are still showing up

lukeholder:
@boboldehampsink we have decided that throwing an exception when trying to save a line item or adjustment whose ID is already deleted from the database shouldn't throw an exception, and instead will log it and return false from the `\craft\commerce\services\OrderAdjustments::saveOrderAdjustment` and `\craft\commerce\services\LineItems::saveLineItem` methods.

We *think* rapid re-saves of the order is causing this but since nowhere in our code do we look at the return value of those saves (and don't run validation) its safe to return false. We could look to add a validation error to the `id` attribute if that is something people would like, but I doubt developers using these methods are saving line items that don't exist often.

Can you please retest on this branch and let me know if the errors have stopped. (You will see messages in the logs but only as information, the same as saving an invalid line item or adjustment).

As for the race condition on order save, we are looking to other solutions but will do so in a larger release of commerce.

lukeholder:
Replaced by https://github.com/craftcms/commerce/pull/4030 which will be merged into 5.4 soon.

PR #3902, opened 2025-02-20
Fix caching bug when calculating available discounts

lukeholder:


lukeholder:
To get this PR fix early, change your `craftcms/commerce` requirement in composer.json to:

```json
"require": {
  "craftcms/commerce": "dev-feature/pt-2457-5x-discount-dont-work-if-custom-price-calculation-for as 5.3.3",
  "...": "..."
}
```

Then run `composer update`.

We will update this PR once the release is out that includes this PR.

src/services/Discounts.php
```diff
@@ -330,12 +343,12 @@ public function getAllActiveDiscounts(?Order $order = null): array
 
         // Pre-qualify discounts based on purchase total
         if ($order) {
-            if ($order->getEmail()) {
+            if ($orderEmail) {
```
nfourtythree: Should this check if `$orderEmail !== '*'` based on the code further up? Either that or leave this code as `$order->getEmail()`

nfourtythree: Although I suppose it is already checking the `$order` variable so should be fine

PR #3908, opened 2025-02-25
Update generics for ProductQuery and VariantQuery

johnnynotsolucky:
### Description

Updates generic definitions for product and variant queries so that type definitions like this will work:

```php
/**
 * @template TKey of array-key
 * @template TElement of Element
 * @param ElementQuery<TKey, TElement> $query
 */
public function doSomethingWithQuery(ElementQuery $query): void
{
	// ...
}


doSomethingWithQuery(Variant::find());
doSomethingWithQuery(Product::find());
```

### Related issues

N/A


lukeholder:
Thanks!

PR #3914, opened 2025-03-05
[5.4] Add ability to hide commerce navigation items

lukeholder:
### Description

This allows main nav items to be hidden even if the user has permission to see them since the site does not use them. For example donations or subscriptions.

### Related issues

https://github.com/craftcms/commerce/discussions/2374


lukeholder:
Discussed internally and recommend this plugin for now: https://plugins.craftcms.com/cp-nav?craft5

PR #3931, opened 2025-03-12
Fix setting custom fields in variant collection

### LINKED ISSUE #3927: [5.x]: Variants in product fixtures don't have their custom fields saved.
### Description

Variants in product fixtures don't have their custom fields saved. Other attributes are saved.

### Steps to reproduce

Create a "product fixture" that defines variants with custom fields.

```php
<?php

return [
    [
        'typeId' => $this->productTypeIds['product'],
        'title' => 'test',
        'variants' => [
            'new1' => [
                'title' => 'foo',
                'sku' => 'this works',
                'fields' => [
                    'customField' => 'this doesnt',
                ],
            ],
        ],
    ],
];
```

### Expected behavior

Variant is created, sku is filled and customField is filled

### Actual behavior

Variant is created, only sku is filled

### Craft CMS version

5.6.11

### Craft Commerce version

5.3.4

### PHP version

8.3

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

boboldehampsink:
### Description

Now also sets custom field values in variant collection, not only attributes

### Related issues

Fixes #3927

### DIFF
--- a/src/elements/VariantCollection.php
+++ b/src/elements/VariantCollection.php
@@ -35,9 +35,16 @@ class VariantCollection extends ElementCollection
                 continue;
             }
 
+            $attributes = $item;
+
             $item = \Craft::createObject(Variant::class, [
-                'config' => ['attributes' => $item],
+                'config' => ['attributes' => $attributes],
             ]);
+
+            // Set custom fields if provided
+            if (isset($attributes['fields'])) {
+                $item->setFieldValues($attributes['fields']);
+            }
         }
 
         /** @var static $collection */


nfourtythree:
Hi @boboldehampsink 

Thank you for bringing this to our attention and taking the time to create a PR we really appreciate it.

I am going to close this PR in favour of the fix in PR #3946

Thanks!

nfourtythree:
The fix in PR #3946 has now been released in Commerce 5.3.6.

Thanks!

boboldehampsink:
This release has other issues for me. Now this is broken:

```php
        $variant = [
            'relatedTo' => [
                'targetElement' => [EntryQuery::find()],
                'field' => 'products',
            ],
            'orderBy' => 'isDefault DESC',
        ];

        return Product::find()
            ->hasVariant($variant)
           ->with([
                ['variants', [
                    ...$variant,
                    'limit' => 1,
                ]],
            ])->all();
```

Could this have anything to do with this fix? 5.3.5 works fine.

nfourtythree:
Hi @boboldehampsink 

I have just added a test to the suite which eager loads the variants: https://github.com/craftcms/commerce/commit/1df2cad4981bdcb34485e4752b486fb64c04d768

At the moment it looks like this is behaving as expected.

I understand your example was probably partial pseudo-code/simplified, so could you explain what you were trying to do with the query so I can make sure to test it thoroughly. As for one example you have `EntryQuery::find()` which is a method that doesn't exist (assuming you meant `Entry::find()`?)

Thanks!

boboldehampsink:
Oh yeah I meant that, it is indeed pseudo-code. The problem is in the "hasVariant" method, when I remove it I get results. With that in place, no results. And this works fine in 5.3.5.

nfourtythree:
Is it just flipping between the Commerce versions and keeping the same CMS version that shows the difference? Or is the CMS version changing as well?

Are you able to xdebug or use the debug toolbar to extract the raw query to see what the difference is?

[Between 5.3.5 and 5.3.6](https://github.com/craftcms/commerce/compare/5.3.5...5.3.6) we didn't make any changes specifically to querying so in theory there shouldn't be a difference. But I am keen to get to the bottom of the issue you are seeing

boboldehampsink:
Only changing Commerce versions. I will look into it some more.

boboldehampsink:
It seems to be in my fixtures. I changed

```
'variants' => [
  ...
  'fields' => [
     'products' => [123],
  ],
];
```

To 

```
'variants' => [
  ...
  'field:products' => [123],
];
```

As required by this change, but it doesn't work.

nfourtythree:
And, to follow along with this, you have a `craft\commerce\fields\Products` custom field on your variant field layout?

Would it be possible to send your tests directory to support@craftcms.com? That might be the easiest way for us to trace through what could be happening here.

boboldehampsink:
I sent you a mail 👍

PR #3933, opened 2025-03-12
Fix upgrading from 4.x with OrderCurrencyValuesAttributeConditionRule in place

boboldehampsink:
### Description

OrderCurrencyValuesAttributeConditionRule exists in 4.x but in 5.x relies on specific 5.x code. When this is in place, upgrading from 4.x to 5.x fails. This PR fixes that.


nfourtythree:
Hi @boboldehampsink 

Thank you for this, could you provide a stack trace of the error?

We are trying to hunt down the origin of what would cause this issue and require the suggested fix.

Thanks!

boboldehampsink:
Don't have it anymore but I had an order field layout with an order conditional in 4.x and tried upgrading to 5.x

boboldehampsink:
@nfourtythree any news on this?

nfourtythree:
Hi @boboldehampsink 

To double check we can do the correct testing. You have field on the order layout setup something like this?

CleanShot 2025-05-06 at 14 18 12@2x

Is that correct?

boboldehampsink:
I have two fields. On is called "original order" where you can select a different order. And one field is called "Order status", which has the conditional that "original order" "has value".

nfourtythree:
Thank you for the fix @boboldehampsink will merge this into `5.x` now and will be in the next release of Commerce.

To get this early, change your `craftcms/commerce` requirement in your project's `composer.json` to:

```json
"require": {
  "craftcms/commerce": "5.x-dev as 5.3.11",
  "...": "..."
}
```

Then run `composer update`.

Thanks!

nfourtythree:
Commerce `5.3.12` has been released with this fix included.

Thanks!

PR #3946, opened 2025-03-17
[5.x] Fix custom field mass assignment in `VariantCollection`

nfourtythree:
### Description
Trying to assign values to custom fields when making a `VariantCollection` could cause the data to be dropped.

This fix will ensure the custom field data is set on the element when passing in the data using one of the following methods (where `myCustomField` is the handle of the custom field:

```php
// Top level
$variant = [
    'title' => 'My Variant',
    // ...
    'myCustomField' => 'Foo',
];

// Prefixed
$variant = [
    'title' => 'My Variant',
    // ...
    'field:myCustomField' => 'Foo',
];

// Nested
$variant = [
    'title' => 'My Variant',
    // ...
    'fieldValues' => [
        'myCustomField' => 'Foo',
    ],
];
```

### Related issues
#3931
#3927


lukeholder (approved):
LGTM!

PR #3970, opened 2025-04-09
Fix: Plugin::$instance->currencies->numericCodeFor should return numeric code

d--j:
Obviously, a Copy & Paste error


lukeholder:
Thanks!

PR #3975, opened 2025-04-11
Fix missing hydration of metadata for pricing rule

wanze:
### Description

The `CatalogPricing` model offers a way to get and set metadata. When saving, the metadata is written to the database. But when loading the model, the metadata is currently missing. This PR adds the missing `metadata` column when loading the data from the database.


lukeholder:
Thanks for this, going with #3982, which will be in Commerce 5.4 once that is released.

lukeholder:
Craft Commerce 5.4 is now out with this fix/improvement. Thanks!

https://plugins.craftcms.com/commerce/changelog?craft5

PR #3995, opened 2025-05-01
Example template - fix list of transactions for offsite gateways and style failed transactions

vnali:
### Description
In the example templates, the order's transactions are listed based on the results of `order.transactions`. For offsite gateways, this may not always work correctly, and there is a possibility that a child transaction is listed below another parent transaction. This PR aims to ensure that child transactions are listed below their related parent transaction. another improvement is failed transactions are styled in red.

### DIFF
--- a/example-templates/src/shop/customer/order.twig
+++ b/example-templates/src/shop/customer/order.twig
@@ -285,27 +285,50 @@
         </h3>
         <table class="w-full">
           <tbody>
-          {% for transaction in order.transactions %}
-            <tr>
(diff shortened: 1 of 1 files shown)


vnali:
Hi
Actually this PR might not work because a child transaction -a success for redirected transaction- can also have its own child transaction, such as refund.

lukeholder:
Thanks for the PR, but we will leave this as is for now, and we will look at improving the example templates in the future. Thanks!

Issue #3571, opened 2024-07-09
[5.x]: New products/variants are not promotable by default


mikesnoeren:
### Description
As per [the documentation](https://craftcms.com/docs/commerce/5.x/system/discounts.html#discount-matching-items), the promotable option should be enabled by default. In Craft 5 + Commerce 5 this is not the case.

### Steps to reproduce

1.Make a product
2.See see the lightswitch is not flipped by default.

### Craft CMS version

 5.2.5

### Craft Commerce version

5.0.11.1

### Installed plugins and versions

-

nfourtythree:
Hi @mikesnoeren 

Thank you for reporting this. We have made an [issue in the docs](https://github.com/craftcms/docs/issues/631) to be updated to make sure they explain how the system is working.

We have also added the `promotable` property to the options for showing in the variant index. This will allow you to have better visibility over which variants have been set to be profitable. This change will be included in the next release of Commerce.

We will take the feedback on board and see if there is a way (maybe a setting) to change the behaviour for whether a new variant is promotable by default.

Thanks!

mikesnoeren:
Hi @nfourtythree, thanks for creating the issue in the docs.

From my perspective, enabling the promotable property by default would be the most effective approach. Any necessary (conditional) rules for promotions should, in my opinion, be managed within the promotion itself.

Introducing this setting seems to add unnecessary complexity to the already intricate promotions functionality. 

Additionally, I find it quite unconventional to restrict certain products from being eligible for promotions. Having worked on multiple webshops, we have never encountered a need for this restriction.

nfourtythree:
Hi @mikesnoeren 

Again, thank you for this feedback. We will take it on board when making decisions regarding this functionality.

Thanks!

philipboomy:
My editors are also asking for "Promotable" to be on as default. 99.5% of the time they create a product/variant it is promotable.

lukeholder:
Craft Commerce 5.4 is now out with this fix/improvement. You can now edit the field settings in the layout to set the default value!

 Thanks!

https://plugins.craftcms.com/commerce/changelog?craft5

PR #4025, opened 2025-05-28
[5.4] Add default options to variant fields

lukeholder:
### Description

- [x] Allow the `promotable` option to have a default in field layout settings.
- [x] Default "Available for purchasable"
- [x] Default "Stock/Inventory tracked" with a default "Allow Out Of Stock Purchases" options too.

### Related issues

#3571

### DIFF
--- a/src/fieldlayoutelements/PurchasablePromotableField.php
+++ b/src/fieldlayoutelements/PurchasablePromotableField.php
@@ -59,7 +59,7 @@ class PurchasablePromotableField extends BaseNativeField
             'id' => 'promotable',
             'name' => 'promotable',
             'small' => true,
-            'on' => $element->getIsDraft() ? $this->defaultPromotable : $element->promotable,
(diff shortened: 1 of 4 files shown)


src/fieldlayoutelements/PurchasablePromotableField.php
```diff
@@ -54,11 +59,23 @@ public function inputHtml(ElementInterface $element = null, bool $static = false
             'id' => 'promotable',
             'name' => 'promotable',
             'small' => true,
-            'on' => $element->promotable,
+            'on' => $element->getIsDraft() ? $this->defaultPromotable : $element->promotable,
```
lukeholder: Not sure if `getIsDraft` is enough to know if its the first time viewing the element. Anything else we could check to know?

lukeholder: Changed it to `.isFresh()`

nfourtythree (approved):
LGTM!

PR #4042, opened 2025-06-11
[5.5] Fixed emails not rendering in the site the order was placed in

nfourtythree:
### Description

When a user places an order on a non-primary site, the confirmation email pulls in content from global fields in the primary site instead of the current site.


CHANGELOG.md
```diff
@@ -4,7 +4,11 @@
 
 - Fixed a bug where purchasable cache was not cleared when stock was updated.
 - Fixed a PHP error that could occur when sending emails. ([#4017](https://github.com/craftcms/commerce/issues/4017))
+<<<<<<< bugfix/send-email-with-current-site-set-to-order-site
```
lukeholder: Bad merge

lukeholder (changes requested):
Looks like it just needs a cleanup, but the fix looks OK to me.

PR #4051, opened 2025-06-18
[5.x] Fix duplicate order reference generation on order complete

lukeholder:
### Description

This just ensures that the same order reference cant be created on order complete using the reference format. It doesn't add a global unique validator to the attribute, and no unique constraint on the column.


nfourtythree (approved):
LGTM 👍

PR #4058, opened 2025-06-23
Fix PHP version check

boboldehampsink:
8.2.0 is actually 80200


boboldehampsink:
cc @nfourtythree

lukeholder:
Thanks, this is now in 5.x

Issue #4090, opened 2025-08-05
[5.x]: Product duplication and variant title not updating with Variant Title Format enabled


andreapasotti:
### Description

We have active a Variant Title Format of this type:
{typologyPurchasable.label|title} {product.startDate|date('d/m/Y H:i')}

We have never encountered problems like this in the past (the site is 2 years old and started with Craft Commerce 4), but for some time now we have noticed that by duplicating a product, changing the “startDate” and saving the product, the variants do not update their title, thus remaining with the “startDate” of the starting product.

### Steps to reproduce

1. Create a product with a custom field.
2. Activate the "Variant Title Format" with that custom field.
3. Fill a product and a variant.
4. Duplicate the product.
5. Update the custom field on the product and save.

### Expected behavior

The variants update their title when the product is saved.

### Actual behavior

The variants' titles are not updated when the product is saved.

### Craft CMS version

5.7.11

### Craft Commerce version

5.4.4

### PHP version

8.2.26

### Operating system and version

Linux 6.12.5-linuxkit / Linux 5.2.0

### Database type and version

MariaDB 10.4.34 / MariaDB 10.11.13

### Image driver and version

Imagick 3.8.0 (ImageMagick 6.9.11-60)

### Installed plugins and versions

- Blitz	5.11.5
- Craft Commerce	5.4.4
- Customizable Radio Buttons Field	2.0.0
- Expanded Singles	3.0.2
- Field Manager	4.0.3
- Gift Voucher	4.0.7
- Guest Entries	4.0.1
- Icon Picker	3.0.4
- ImageOptimize	5.0.7
- Maps	5.0.4
- Mix	v1.8.1
- oEmbed	3.1.5
- PayPal Checkout for Craft Commerce	3.0.3
- Redactor	4.2.0
- Related Elements	1.2.0
- SEOmatic	5.1.16
- Simple Mailchimp	3.0.1
- Snaptcha	5.1.1
- Sprig	3.6.1
- Stripe for Craft Commerce	5.0.7
- Super Table	4.0.4
- Twig Perversion	5.0.1
- Typed link field	3.0.0-beta
- Typogrify	5.0.1
- Wheel Form	4.0.3

lukeholder:
Thanks for reporting this. I see the issue.

Since the variants are not considered 'dirty' (unchanged) during the product save, they are not being re-saved.

I will look into a fix. Thanks.

andreapasotti:
Thank You @lukeholder !

[...]

PR #4095, opened 2025-08-13
Set variants to dirty if the variantTitleFormat references products

lukeholder:
### Description

Possible brute force solution to #4090 - for discussion.


nfourtythree (approved):
Until there is a solution in Craft for this issue with nested elements I think this looks good to me!

Issue #4089, opened 2025-08-01
[5.x]: Cart cleared after registering new user during checkout


bennonite:
### Description
When I register a new account during the checkout flow the active cart gets cleared. This sounds perhaps related to [4024](https://github.com/craftcms/commerce/issues/4024) and when I downgrade Craft to 5.3.9, like they did, the checkout flow works as expected - however, unlike that issue I'm not using any custom modules to update the cart email. I do have users set to require email verification, and if the guest does verify their email, logs in, and goes back to the cart it's there. 

I brought in the template shop and am experiencing the issue there too: 

https://github.com/user-attachments/assets/1c76cb31-5f8f-477d-b778-4d3c2e0e446b

### Steps to reproduce

1. Add products to cart (as guest)
2. Enter new email address
3. Register a new account
4. Cart is missing on next page (addresses)

### Craft CMS version

5.8.12

### Craft Commerce version

5.4.4

### PHP version

8.2.28

### Operating system and version

Linux 6.10.14-linuxkit

### Database type and version

MySQL 8.0.40

### Image driver and version

Imagick 3.7.0 (ImageMagick 6.9.11-60)

### Installed plugins and versions

-

lukeholder:
@bennonite Thanks for reporting this. I can reproduce the bug. I am looking into a fix.

In the meantime could using the [registerUserOnOrderComplete](https://craftcms.com/docs/commerce/5.x/reference/controller-actions.html#supported-params-3) param on the cart update form be a solution for you?

bennonite:
Thanks @lukeholder! This site isn't live yet so I can wait for the fix, though great to know there's a work around available here too

philipboomy:
@lukeholder I am running into an issue where when I share a cart its not "creating" the cart in the customer's browser. Could this be related to the above issue? Been trying to troubleshoot but so far no luck.

update: Looks like its related. If I use an email address that already exist for an order and then share it, it will not create the order in the browser. If I use an email that is completely new it will create it in the browser.

lukeholder:
@philipboomy Yes sounds like it is related. We are working on a fix and will update soon.

philipboomy:
Any update on this? Thanks

lukeholder:
Working on a couple fixes and will likely be in the release tonight. Thanks.

lukeholder:
Thanks for reporting this. Here’s the PR with a fix for registering during checkout: https://github.com/craftcms/commerce/pull/4102

[...]

PR #4102, opened 2025-08-20
Prevents the loss of cart when registering during checkout

lukeholder:
### Description

When a customer creates a credentialed user by registering, we need to allow them to keep the credentialed customers cart.

### Related issues

#4089

### DIFF
--- a/src/services/Carts.php
+++ b/src/services/Carts.php
@@ -549,11 +549,12 @@ class Carts extends Component
     {
         $segments = Craft::$app->getRequest()->getActionSegments();
         $userSaveSegments = ['users', 'save-user'];
+        $isUserSaveAction = $segments == $userSaveSegments;
 
         // we have a cart number, currently anon, and the current action being executed is user save
         if (!Craft::$app->getUser()->getIdentity() &&
             !Craft::$app->getRequest()->getIsCpRequest() &&
-            sort($segments) == sort($userSaveSegments)
+            $isUserSaveAction
         ) {
             $currentCartNumber = $this->getSessionCartNumber();
             // Set the session flag to preserve the cart for this user
--- a/src/services/Carts.php
+++ b/src/services/Carts.php
@@ -13,6 +13,7 @@ use craft\commerce\elements\Order;
 use craft\commerce\events\CartPurgeEvent;
 use craft\commerce\Plugin;
 use craft\db\Query;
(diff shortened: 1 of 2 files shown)


src/services/Carts.php
```diff
@@ -537,4 +538,27 @@ private function _getCartPaymentCurrencyIso(): string
...
         return Plugin::getInstance()->getPaymentCurrencies()->getPrimaryPaymentCurrencyIso();
     }
+
+    /**
+     * @param User $user
+     * @return void
+     * @throws MissingComponentException
+     * @throws Throwable
+     */
+    public function customerRegisteredHandler(User $user): void
```
nfourtythree: Should we name this similar to the other handlers? Just in case we want to use it for more than one purpose in the future?

`afterSaveUserHandler` would be my suggestion

src/services/Carts.php
```diff
@@ -537,4 +538,27 @@ private function _getCartPaymentCurrencyIso(): string
...
+    {
+        $segments = Craft::$app->getRequest()->getActionSegments();
+        $userSaveSegments = ['users', 'save-user'];
+        $isUserSaveAction = $segments == $userSaveSegments;
+
+        // we have a cart number, currently anon, and the current action being executed is user save
+        if (!Craft::$app->getUser()->getIdentity() &&
```
nfourtythree, line 556: Do we need an extra check in here for if this is a new user? For example this will be called if a user saves their profile on the front end, does that have an impact on this code e.g. should it run?

src/services/Carts.php
```diff
@@ -537,4 +538,27 @@ private function _getCartPaymentCurrencyIso(): string
...
+            !Craft::$app->getRequest()->getIsCpRequest() &&
+            $isUserSaveAction
+        ) {
+            $currentCartNumber = $this->getSessionCartNumber();
+            // Set the session flag to preserve the cart for this user
+            Craft::$app->getSession()->set('commerce:anonymousCartWithCredentialedCustomer:' . $currentCartNumber, true);
```
nfourtythree, line 562: This key is used in a few places now, should we make a common method or constant to retrieve it?

nfourtythree (changes requested):
Logic looks sound, I believe this is what we need to tackle the issue. Have just made a couple of very minor notes.

PR #4116, opened 2025-09-08
[5.5] Adds a force option to reset-data command

lukeholder:
### Description

Adds a `--force` to the data reset command to avoid confirmation prompt.

### Related issues

https://github.com/craftcms/commerce/discussions/4115


lukeholder:
Reminder: add change-log entry before merging.

PR #4129, opened 2025-10-03
Bugfix/4125 variant duplication and deletion

i-just:
### Description
**issue 1** (described in the original report):
When duplicating a variant, the duplicate was only created against the canonical product and not the product’s provisional draft.

Fixed by relying on the nested element trait eager loading map logic for `owner` and `primaryOwner`.

**issue 2**:
When you delete a variant, it appears to be deleted in the provisional draft, but when you attempt to fully save the product, the deleted variant comes back.
(This only started happening after implementing a fix for the first issue.)

Fixed, by adjusting the change we made here: https://github.com/craftcms/commerce/commit/fc5564717396d3662927effb2baf56e957e42d2b

@nfourtythree, feel free to adjust it in any way you like. Also, LMK if you’d like to chat.

### Related issues
#4125

PR #4133, opened 2025-10-09
[5.x] Variant eager loading - only array merge for owner and primaryOwner

i-just:
### Description
Part 2 of the https://github.com/craftcms/commerce/pull/4129.

We only want to merge in the `elementType` for the `owner` and `primaryOwner` handles.


nfourtythree (approved):
As discussed, think this covers the bases and models the underlying trait's method

PR #4193, opened 2025-12-08
[5.x] Add view column options to inventory screen

lukeholder:
### Description

Work in Progress.

Adds view options to the inventory screen to show/hide columns. Options are stored in local storage like the element index.

CleanShot 2025-12-08 at 17 38 25@2x

Uses the same UI as an element index.

TODO:

- [x] Add and check all new translations

### Related issues

https://github.com/craftcms/commerce/discussions/3520
https://github.com/craftcms/commerce/discussions/3429


lukeholder:
Commerce 5.7 is now out with this included. Thanks!

PR #4203, opened 2026-01-15
Protect against loading cart belonging to credentialed users

lukeholder:
### Description

Carts should only be loadable if not associated with a non-credentialed user - Or the current user is the same user as the credentialed customer associated with the cart.


lukeholder:
Replaced by #4207

Issue #4208, opened 2026-01-26
[5.x]: Tax adjustments get overwritten as 'discount' in stead of 'tax'


joostwaaijer:
### Description
We have a shop where we show all prices including VAT, when a customer inputs a valid tax id, the customer should be exempt from paying taxes. This currently shows as a discount adjustment instead of a tax adjustment.

This seems wrong to me, because it's technically not a discount that you give. The problem gets bigger when a customer also adds a coupon code that discounts lineItems in their cart, we currently can't force the correct order, so that the discount gets calculated after removing taxes. When I have a look in the code the adjustment type gets overwritten here:  https://github.com/craftcms/commerce/blame/5.x/src/adjusters/Tax.php#L244 while the file itself has a constant ADJUSTMENT_TYPE, that gets set in the `_createAdjustment`

### Steps to reproduce

1. Create a VAT rule for a certain country, including in the price checked
2. Create a coupon code that gives discount
3. Create an order with the coupon code and a valid tax id for the created vat rule
4. See that the removal of VAT is seen as a Discount adjustment instead of a Tax adjustment

### Expected behavior
I expect the VAT removal on the order would be shown as a tax adjustment

### Actual behavior
The VAT removal on the order is seen as a discount adjustment:

<img width="2563" height="843" alt="Image" src="https://github.com/user-attachments/assets/a29d68c8-fd95-412f-beb9-b0e4dd16fa49" />

### Craft CMS version

Craft Pro 5.8.23

### Craft Commerce version

5.5.2

### PHP version

8.4.17

### Database type and version

PostgreSQL 17.5

### Installed plugins and versions

-

joostwaaijer:
I've created a Pull Request that sets the adjustment type correctly in the following PR: https://github.com/craftcms/commerce/pull/4209

lukeholder:
Hi @joostwaaijer,

Thanks for the PR.

Unfortunately, at the moment, this wouldn’t make sense given how Commerce currently works.

It would cause incorrect total tax calculations and also break other cart totals, resulting in an inaccurate cart.

We’ve been removing the value of tax included with a “discount” since Commerce 1.x, so this isn’t something we can change so simply. The tax adjuster system runs last, and we don’t remove the amount from the price directly at the end of the process, as this would break the discount and shipping calculations that run before it, leading to incoherent cart totals.

[...]

PR #4209, opened 2026-01-26
Fix tax adjustment type

joostwaaijer:
### Description
Makes the tax adjustment type `tax` instead of `discount`

### Related issues
https://github.com/craftcms/commerce/issues/4208

### DIFF
--- a/src/adjusters/Tax.php
+++ b/src/adjusters/Tax.php
@@ -192,7 +192,7 @@ class Tax extends Component implements AdjusterInterface
                 // We need to display the adjustment that removed the included tax
                 $adjustment->name = Craft::t('site', $taxRate->name) . ' ' . Craft::t('commerce', 'Removed');
                 $adjustment->amount = $orderLevelAmountToBeRemovedByDiscount;
-                $adjustment->type = 'discount'; // TODO Not use a discount adjustment, but modify the price of the item instead. #COM-26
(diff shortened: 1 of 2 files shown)


lukeholder:
Hi @joostwaaijer,

Thanks for the PR.

Unfortunately, at the moment, this wouldn’t make sense given how Commerce currently works.

It would cause incorrect total tax calculations and also break other cart totals, resulting in an inaccurate cart.

We’ve been removing the value of tax included with a “discount” since Commerce 1.x, so this isn’t something we can change so simply. The tax adjuster system runs last, and we don’t remove the amount from the price directly at the end of the process, as this would break the discount and shipping calculations that run before it, leading to incoherent cart totals.

That said, we do have this on the list to explore whether we can add a setting or option to remove included tax from the price during line item population (before adjustments run). However, this would require a much larger change and careful consideration. We’d need to work on this on our side to ensure all integrations and extensions (such as gateways) can support both modes.

If showing the removed included tax as a discount doesn’t work for your client, we recommend entering prices without tax and displaying any default included tax using your own front-end calculation. This simplifies tax handling, as no negative price adjustment is required.

I’ve created a discussion here (https://github.com/craftcms/commerce/discussions/3876) and would love your input and client requirements. Thanks!

joostwaaijer:
@lukeholder Thanks for the explanation! I'll take this to our client and see if we can work with the proposed solution. I'll add my input to the discussion!

Issue #4199, opened 2026-01-12
[5.x]: Error when merging canonical changes to variants into draft


elivz:
### Description

Very occasionally we get into a situation where the canonical version of a product has changes to the variants, relative to a particular user's draft version. When Craft tries to merge those changes into the draft, it errors out due to the product's `allVariants` attribute not having a setter. I understand that `allVariants` is a stopgap and will be going away again at some point in the future, but in the mean time I think it either needs a setter, or to be excluded from the `mergeCanonicalChanges` routine somehow.

### Stack trace

```
yii\base\InvalidCallException: Setting read-only property: craft\commerce\elements\Product::allVariants in /var/www/vendor/yiisoft/yii2/base/Component.php:218
Stack trace:
#0 /var/www/vendor/craftcms/cms/src/base/Element.php(2664): yii\base\Component->__set()
#1 /var/www/vendor/craftcms/cms/src/base/Element.php(3328): craft\base\Element->__set()
#2 /var/www/vendor/craftcms/cms/src/services/Elements.php(1425): craft\base\Element->mergeCanonicalChanges()
#3 [internal function]: craft\services\Elements->{closure:{closure:craft\services\Elements::mergeCanonicalChanges():1405}:1406}()
#4 /var/www/vendor/yiisoft/yii2/db/Connection.php(818): call_user_func()
#5 /var/www/vendor/craftcms/cms/src/services/Elements.php(1406): yii\db\Connection->transaction()
#6 /var/www/vendor/craftcms/cms/src/services/Elements.php(1253): craft\services\Elements->{closure:craft\services\Elements::mergeCanonicalChanges():1405}()
#7 /var/www/vendor/craftcms/cms/src/services/Elements.php(1405): craft\services\Elements->ensureBulkOp()
#8 /var/www/vendor/craftcms/cms/src/controllers/ElementsController.php(296): craft\services\Elements->mergeCanonicalChanges()
#9 [internal function]: craft\controllers\ElementsController->actionEdit()
#10 /var/www/vendor/yiisoft/yii2/base/InlineAction.php(57): call_user_func_array()
#11 /var/www/vendor/yiisoft/yii2/base/Controller.php(178): yii\base\InlineAction->runWithParams()
#12 /var/www/vendor/yiisoft/yii2/base/Module.php(552): yii\base\Controller->runAction()
#13 /var/www/vendor/craftcms/cms/src/web/Application.php(360): yii\base\Module->runAction()
[...]

PR #4222, opened 2026-02-05
Product/variant management improvements

brandonkelly:
Switches the variant manager’s attribute from `allVariants` to `variants`, and configures it with a `valueGetter` that returns `$this->getVariants(true)`. That fixes a couple issues:

- Fixes a potential error when merging canonical product changes into a draft (#4199)
- Fixes a bug where variants weren’t being shown as modified when variants were added/deleted/reordered (since `VariantsField::$attribute` is `variants`).

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -62,25 +62,25 @@
         },
         {
             "name": "brick/math",
-            "version": "0.12.3",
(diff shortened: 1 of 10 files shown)


lukeholder:
🎉

PR #4244, opened 2026-03-09
[5.x] fix(manual-gateway): Fixed a PHP error that could occur when getOnlyAllowForZeroPriceOrders was not set. 

john-henry:
Ensures boolean parsing defaults to false when null for the zero price order condition.

### Description
TypeError: craft\commerce\gateways\Manual::getOnlyAllowForZeroPriceOrders(): Return value must be of type string|bool, null returned


lukeholder:
This will be merged into 5.x when 4.11 releases

PR #4245, opened 2026-03-09
[4.x] fix(manual-gateway): Fixed a PHP error that could occur when getOnlyAllowForZeroPriceOrders was not set

john-henry:
Ensures boolean env parsing fallback to false to prevent PHP errors in Manual Gateway.

### Description

TypeError: craft\commerce\gateways\Manual::getOnlyAllowForZeroPriceOrders(): Return value must be of type string|bool, null returned


brandonkelly:
Thanks for the PR! We just tagged Commerce 4.11 with this change.

PR #4246, opened 2026-03-09
Add read-only mode to Carts::getCart()

boboldehampsink:
## Why
Calling craft.commerce.carts.getCart() for display-only UI (like a header cart badge) currently initializes the cart session and sets a cart cookie when none exists. On Cloudflare-cached pages this creates avoidable cache fragmentation and unnecessary Set-Cookie responses.

This change adds a read-only retrieval path so templates/controllers can check for an existing cart without starting a new cart session.

## What changed
- Added readOnly parameter to Carts::getCart()
- getCart(readOnly: true):
  - returns the existing cart if one is already associated with the request cookie
  - returns null when no cart cookie/cart exists
  - does not create a new cart number
  - does not call setSessionCartNumber()
- Added a unit test that asserts read-only mode does not write session cart state

## Example use cases
Use read-only mode when you need to read cart state but must avoid mutating session/cookie state:

PHP example:
$cart = Plugin::getInstance()->getCarts()->getCart(readOnly: true);
$count = $cart?->totalQty ?? 0;

Twig example:
{% set cart = craft.commerce.carts.getCart(false, true) %}
{{ cart ? cart.totalQty : 0 }}

Typical scenarios:
- Cached global header cart count on product/category/marketing pages
- Edge-cached storefront pages where Set-Cookie should be avoided for anonymous traffic
- Read-only telemetry or analytics that checks current cart presence

Use regular getCart() for cart/checkout flows that are expected to create or mutate cart state.

### DIFF
--- a/src/services/Carts.php
+++ b/src/services/Carts.php
@@ -128,12 +128,33 @@ class Carts extends Component
      * Get the current cart for this session.
      *
      * @param bool $forceSave Force the cart.
+     * @param bool $readOnly Whether to retrieve the cart in read-only mode.
      * @throws ElementNotFoundException
      * @throws Exception
      * @throws Throwable
      */
-    public function getCart(bool $forceSave = false): Order
+    public function getCart(bool $forceSave = false, bool $readOnly = false): ?Order
     {
+        if ($readOnly) {
+            if (isset($this->_cart)) {
+                return $this->_cart;
+            }
+
+            if (!$this->getHasSessionCartNumber()) {
+                return null;
+            }
+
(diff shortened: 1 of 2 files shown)


lukeholder:
Thanks, some form of this will be in 5.7

lukeholder:
We will work on this from https://github.com/craftcms/commerce/pull/4263 Thanks

lukeholder:
`peek` param was added to `commerce/cart/get-cart` action in Commerce 5.7

PR #4259, opened 2026-03-20
Include billing and shipping address conditions in Gateway edit form

johnnynotsolucky:
### Description

We have a need to include gateways on orders by address. The billing and shipping address conditions already exist on the Gateway class, they're just not being rendered in the edit form, or included in the save action.

This PR updates the edit form to show both condition inputs, and updates the save action so that they can be persisted correctly.

CleanShot 2026-03-20 at 13 57 32


lukeholder:
Thankyou, we have included a fix like this this in the next release.

```json
"require": {
  "craftcms/commerce": "5.x-dev#15ddcbde0e333d0b79bba0178bb8425275afbaca as 5.6.0",
  "...": "..."
}
```

Then run `composer update`.

We will update this PR once the release is out.

lukeholder:
Commerce 5.6.1 is now out with this!

PR #4290, opened 2026-04-29
[5.7] User deletion blockers

nfourtythree:
### Description

Craft 5.10 introduced a structured deletion blocker system that lets plugins register actionable blockers when an element can't be deleted, rather than silently failing or throwing a generic error.

This PR migrates Commerce's existing user-deletion protection to use that system. Previously, attempting to delete a user with orders would silently prevent deletion, and attempting to delete a user with subscriptions would throw an unhandled exception. Now, store admins are presented with clear explanations of what is blocking deletion and concrete actions to resolve each blocker before proceeding.

### Orders blocker

When a user has completed orders, an `OrderCustomersDeletionBlocker` is registered. The admin can choose to:

- **Reassign orders** — pick a different user to become the customer on all affected orders
- **Remove customer data** — nullify the customer association and email from orders, with optional removal of billing address, shipping address, and completed email
- **Delete orders** — permanently delete the orders via the standard element deletion flow

Orders that are reassigned or stripped of customer data remain in the system. A new `customerDeleted` boolean column on the orders table tracks whether customer data was removed, and the order edit screen shows a notice when this is the case.

### Subscriptions blocker

When a user has subscriptions, a `SubscriptionCustomersDeletionBlocker` is registered per gateway (since cancellation behaviour differs per gateway). The admin can choose to:

- **Delete subscriptions** — optionally cancel with the gateway first, then hard-delete the subscription elements

Subscriptions are also now cascade-deleted at the database level when a user is hard-deleted, so if the admin proceeds past the blocker without explicitly deleting subscriptions, the database FK (`SET CASCADE`) handles cleanup automatically. A migration updates the FK from `RESTRICT` to `CASCADE`.

### Other changes

- `Subscription::getSubscriber()` now correctly returns `?User` (it could already return `null` in practice but wasn't typed as such)
- Requires Craft CMS 5.10.0 or later

---

### Extensibility

- Added `craft\commerce\controllers\OrdersController::actionReassign()`.
- Added `craft\commerce\controllers\OrdersController::actionReassignModal()`.
- Added `craft\commerce\controllers\OrdersController::actionRemoveCustomerData()`.
- Added `craft\commerce\controllers\OrdersController::actionRemoveCustomerDataModal()`.
- Added `craft\commerce\controllers\SubscriptionsController::actionDeleteSubscriptions()`.
- Added `craft\commerce\controllers\SubscriptionsController::actionDeleteSubscriptionsModal()`.
- Added `craft\commerce\elements\Order::getCustomerDeleted()`.
- Added `craft\commerce\elements\Order::setCustomerDeleted()`.
- Added `craft\commerce\elements\deletionblockers\OrderCustomersDeletionBlocker`.
- Added `craft\commerce\elements\deletionblockers\SubscriptionCustomersDeletionBlocker`.
- Added `craft\commerce\services\Orders::reassignOrders()`.
- Added `craft\commerce\services\Orders::removeCustomerData()`.
- `craft\commerce\elements\Subscription::getSubscriber()` now returns `?User` instead of `User`.

### System

- Craft Commerce now requires Craft CMS 5.10.0 or later.
- When deleting a user with orders or subscriptions, store admins are now presented with actionable options to resolve the blocker (reassign orders, remove customer data, or delete subscriptions), rather than a generic error.
- Subscriptions are now cascade-deleted when their associated user is hard-deleted.


denisyilmaz:
@lukeholder is there an ETA for the 5.7 release? We are launching a project mid June which requires user account deletion while retaining orders (with stripped info). so this feature comes at the right time for us. Would be lovely to know if we can plan with it being released soon.

lukeholder:
@denisyilmaz should be out in the next 2-3 weeks.

PR #4318, opened 2026-06-16
Fixed debug panel triggers cart recalculation against primary site in multi-site setups

### LINKED ISSUE #4317: [5.x]: Multi-site: debug panel triggers cart recalculation against primary site
### Description
In a multi-site / multi-store Craft Commerce setup, a product (or variant) can be disabled on the primary site while remaining enabled and purchasable on a secondary site.

Adding the product to the cart from the secondary site works as expected. However, when the page is refreshed with devMode enabled and the debug toolbar active on the frontend, the cart contents are unexpectedly modified.

The issue appears to originate from \craft\commerce\Plugin::_registerDebugPanels(), which retrieves the current cart while operating under the context of the primary site. During this process, Commerce detects that the cart's site ID differs from the current site ID and triggers a cart recalculation for the primary site.

Because the product/variant is disabled on the primary site, the recalculation removes it from the cart, even though it was legitimately added from the secondary site where it is enabled.

### Steps to reproduce

1. Configure a Craft Commerce installation with multiple sites and stores.
2. Create a product and variant that: is disabled on the primary site and is enabled on a secondary site.
3. Visit the secondary site.
4. Enable devMode and ensure the debug toolbar is enabled on the frontend.
5. Add the product to the cart.
6. Verify that the product is present in the cart.
7. Refresh the page.

### Expected behavior
The cart should remain unchanged after the page refresh. Products added from the secondary site should stay in the cart regardless of whether the debug toolbar is enabled.

### Actual behavior
On page refresh, the debug panel retrieves the cart under the primary-site context, causing Commerce to recalculate the cart against the wrong site. As a result, products or variants that are disabled on the primary site are removed from the cart.

### Craft CMS version

5.9.23

### Craft Commerce version

5.6.5

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

lukeholder:
In `_registerDebugPanels()`, the `CommercePanel` was passed the result of `getCart()` eagerly inside an `EVENT_BEFORE_REQUEST` handler. At that point in the request lifecycle, Craft's routing hasn't run yet, so `getCurrentSite()` returns the primary site. `getCart()` detects the mismatch between the cart's `orderSiteId` (a secondary site) and the current site (primary), overwrites `orderSiteId`, and saves — triggering a recalculation under the wrong site that removes line items only enabled on the secondary site.

The fix defers the `getCart()` call from `EVENT_BEFORE_REQUEST` to `CommercePanel::save()`, which Yii2 invokes after the action runs (when the site context is correctly established).

Fixes #4317.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fixed a bug where the debug toolbar could trigger cart recalculation under the primary site context in multi-site setups, causing line items valid only on secondary sites to be removed from the cart. ([#4317](https://github.com/craftcms/commerce/issues/4317))
 - Fixed a bug where the widget Date Range field could break when reopening settings after saving a custom date range. ([#4306](https://github.com/craftcms/commerce/issues/4306))
 - Fixed a bug where new orders created in the control panel would default to a non-primary Order Site in multi-site/single-store setups. ([#4310](https://github.com/craftcms/commerce/issues/4310))
 - Fixed a PHP error that could occur when opening a product with a provisional draft. ([#4314](https://github.com/craftcms/commerce/issues/4314))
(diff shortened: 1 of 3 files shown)


nfourtythree (approved):
Yea, looks good to me!

PR #4343, opened 2026-07-28
Adjust CP request handling for console requests

tommysvr:
### Description

Calling Order::getPdfUrl() from the console throws `'Calling unknown method: craft\console\Request::setIsCpRequest()'` because `setIsCpRequest()` is only defined on `craft\web\Request`, not `craft\console\Request`.

This guards it so `setIsCpRequest` is skipped outside a web request.

Test command to run:

```
public function actionTestCpRequest(): int {
  $order = Order::find()->isCompleted(true)->orderBy('id desc')->one();

  if (!$order) {
    $this->stdout("No completed order found to test with.\n");
    return ExitCode::UNSPECIFIED_ERROR;
  }

  $url = $order->getPdfUrl();
  $this->stdout("URL: $url\n");
  return ExitCode::OK;
}
```

### Related issues
Via Support

### DIFF
--- a/src/services/Carts.php
+++ b/src/services/Carts.php
@@ -457,14 +457,21 @@ class Carts extends Component
 
         $request = Craft::$app->getRequest();
         $isCpRequest = $request->getIsCpRequest();
-        $request->setIsCpRequest(false);
-        $url = UrlHelper::actionUrl('commerce/cart/load-cart', [
-            'number' => $cart->number,
-            'code' => $token,
-        ]);
-        $request->setIsCpRequest($isCpRequest);
 
-        return $url;
+        if ($isCpRequest) {
+            $request->setIsCpRequest(false);
+        }
+
+        try {
+            return UrlHelper::actionUrl('commerce/cart/load-cart', [
+                'number' => $cart->number,
+                'code' => $token,
+            ]);
+        } finally {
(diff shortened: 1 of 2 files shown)


lukeholder:
Thanks for this. I pushed a follow-up commit with two changes:

Simplified the guard instead of checking getIsConsoleRequest() around each call, it now checks if ($isCpRequest) using the value already fetched. 

`craft\console\Request::getIsCpRequest()` is hardcoded to always return false, so this naturally skips the setIsCpRequest() calls on console requests too.

It also skips the no-op setIsCpRequest(false) on ordinary site requests, and wraps the URL generation in try/finally so the flag is restored even if UrlHelper::actionUrl() throws just in case.

Applied the identical fix to Carts::getLoadCartUrl(), which had the same unguarded pattern and is exposed via order.loadCartUrl in templates so it would hit the same fatal error from a queue-sent order email that references that URL.

lukeholder:

```json
"require": {
  "craftcms/commerce": "5.x-dev#20415deaa8352fa1038df42960ea1ccd6e63e09a as 5.7.1",
  "...": "..."
}
```

Then run `composer update`.

nfourtythree:
Commerce `5.7.2` has now been released with this fix included.

Thanks!

PR #4354, opened 2026-09-01
Fix SKU format not resolving {id} for new variants

lukeholder:
## Summary

- `skuFormat` values containing `{id}` produced a truncated SKU (e.g. `SKU-4012-`) for new variants, because `updateSku()` runs before the variant has an ID.
- `updateTitle()` got the same fix for `variantTitleFormat` in #4308, via `afterAssignedId()`. This applies the same idea to SKUs.


nfourtythree (approved):
Looks good to me 👍

Issue #4141, opened 2025-10-15
[5.x]: Multiple queue runners can overlap CatalogicPricing job


boboldehampsink:
### Description

In some cases we use multiple workers that run queue jobs, and CatalogPricing jobs can overlap. In that case (we use Postgres) we get errors like:

```
[94697] Generating catalog pricing. (attempt: 1, pid: 27) - Error (time: 1.159s): SQLSTATE[40P01]: Deadlock detected: 7 ERROR:  deadlock detected
DETAIL:  Process 1074673 waits for ShareLock on transaction 1825032; blocked by process 1074311.
Process 1074311 waits for ExclusiveLock on tuple (109,59) of relation 48920 of database 16385; blocked by process 1074675.
Process 1074675 waits for ShareLock on transaction 1825054; blocked by process 1074673.
HINT:  See server log for query details.
CONTEXT:  while deleting tuple (111,52) in relation "commerce_catalogpricing"
The SQL being executed was: DELETE FROM "commerce_catalogpricing" WHERE "purchasableId" IN (2118, 2119, 2120)
```

### Proposal

Add a mutex lock to this job and make it retryable.

### Craft CMS version

5

### Craft Commerce version

5

### PHP version

8.3

### Database type and version

PostgreSQL 17

### Installed plugins and versions

-

sjcallender:
Similar to/same as https://github.com/craftcms/commerce/issues/4003

jesuismaxime:
Is there somebody on the case, I think we got this issue un our Test environment, leading MySQL to just... crash.

We're unable to create a product because of the auto-save - looks like a product without a variant and/or a variant without a price (draft) causes the overload.

lukeholder:
Hey everyone, can you please make sure you are on the latest releases of Craft and Commerce? We made improvements that reduce the number of calls to rebuild the pricing catalogue. Can you let us know what versions you are on?

For example in 5.4.10 - 2025-11-12, Fixed a bug where duplicate pricing catalog jobs could be queued. (https://github.com/craftcms/commerce/issues/4136)

lukeholder:
Thanks all! Please create a new ticket if the issue persists. We know improvements have been made and are investigating further, but version 5.4.10 has significantly improved the situation.

Issue #4123, opened 2025-09-17
[5.x]: Thousands of Generating catalog pricing jobs


lenvanessen:
### Description
We have a import running at night, since the upgrade that introduced the "Generating catalog pricing job", it taks a loooong time to finish, adding thousands of jobs because we have many products. It disrupts daily operations, where in the past it was finished at night.

Can we have a batch mode/flag/setting to disable the listener? So we can manually batch generate them after the last job finished?

### Craft CMS version

5.x

### Craft Commerce version

4.x

### Installed plugins and versions

-

nfourtythree:
Hi @lenvanessen 

Thank you for your message.

As these queue jobs are being generated from your custom import you have the ability to prevent them from being added and handle the generation yourself.

In you custom code where you are importing your products, just before the saving of the product element, you can prevent the catalog pricing job from being added to the queue.

That can be done with the following code:

```php
Event::once(Queue::class, Queue::EVENT_BEFORE_PUSH, function(PushEvent $event) {
    if ($event->job instanceof CatalogPricing) {
        $event->handled = true;
    }
});

// Save your product
``` 

An important point to note is that by not running the jobs during your sync process it would be required to run the catalog pricing generation after you have finished saving everything. You can do that with the following code:


```php
Plugin::getInstance()->getCatalogPricing()->generateCatalogPrices();
```

Generating the prices all at once at the end shouldn't be too tasking.

We are currently looking into seeing if we can merge pricing jobs together (similar to how search indexing jobs work), so there will be an improvement on this in the future. For that reason I will leave the issue open.

Thanks!

nfourtythree:
Closing this issue for now as solved with the solution above as well as an improvement fix that went out in version `5.4.10`.

That fix prevented duplicate pricing jobs from being added to the queue.

Thanks!

Issue #4003, opened 2025-05-12
[5.x]: Deadlock on Generating catalog pricing job


jgpadua:
### Description
I have been receiving the following Deadlock error quite frequently when Craft executes the "Generating catalog pricing" job. My feeling is that since it started, it has become more recent, but I have not been able to find any logic that indicates that it happened for any specific reason.

`SQLSTATE[40001]: Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction The SQL being executed was: INSERT INTO `commerce_catalogpricing` (`price`, `purchasableId`, `storeId`, `uid`, `dateCreated`, `dateUpdated`) SELECT `basePrice`, `purchasableId`, `storeId`, UUID(), NOW(), NOW() FROM `commerce_purchasables_stores` WHERE `purchasableId` IN (20168)`

Is this a known bug? Is there anything I can do to reduce its occurrence?

### Craft CMS version

5.6.17

### Craft Commerce version

5.3.9

### PHP version

8.3.8

### Operating system and version

Linux 6.8.0-1021-aws

### Database type and version

Mysql 8.0.32

### Installed plugins and versions

-

nfourtythree:
Hi @jgpadua 

Thank you for your message, this is not something we have received reports about.

Catalog prices, by default, are only generated on save of purchasables.

Do you have any custom code that may cause a lot of purchasable saves? Do you see a lot of catalog pricing jobs in your queue?

Thanks!

jgpadua:
@nfourtythree In our case, we have a product sync that happens every day and we also receive updates whenever a product is updated in a third-party tool. In other words, we can update several products at a given time. Each product is added to a queue and the update is executed. Do you have any recommendations to avoid this?

Another question, I'm not sure how this works in Craft, but in our case the products have a fixed price (they don't have a catalog of prices). Is it still necessary for this job to be executed or would it be unnecessary in our case? If unnecessary, would there be a way to skip this job?

![Image](https://github.com/user-attachments/assets/77aa54e4-cfd0-4d3e-8d96-ebba08ff1c9d)

franck-thorck:
Same issue here, we setup other price for variants, but the new prices don't display and the warning Pricing jobs are currently running still displays.

We are on `"craftcms/commerce": "5.3.12"`

Thank you,

nfourtythree:
Hi @jgpadua 

[...]

PR #4360, opened 2026-09-03
[5.x] Fix catalog pricing saves failing on the queue mutex

sjcallender:
This one has been nagging us for a long time. And it's about to bite us again on a new build, so we dove in deeper to offer what we see is a solid fix.

Specifically, the build pulls its products in from an ERP. Running the sync piles up so many catalog pricing jobs that the sync itself starts failing on the mutex. Pausing the queue is the only way to get through it. We see the same thing on another build whenever a lot of products get created at once.

Pausing the queue works for the console command, but we're importing off of webhooks, so the same saves will be running in a web request where we can't pause anything.

[edited]
#4123 was closed asking for a new ticket if it persisted. On 5.7.3 we still get 600 jobs for 600 saves.

This doesn't fix the database deadlocks in #4003 and #4141, which are in `generateCatalogPrices()` rather than the queue table. Fewer jobs means fewer of them running at once, so it should reduce how often those happen.
[/end edit]

## Problem

N saves push N jobs against the one row they all merged into. All but a few reserve nothing, each taking the `catalogpricingqueue` mutex on the way.

The two sides of that mutex disagree:

| | Timeout | On failure |
|---|---|---|
| `reserveCatalogPricingQueueRow()` | `0` | returns `null` |
| `_queueCatalogPricingIds()` | `5` | `throw new RuntimeException` |

So a background job can fail a foreground save.

## Changes

| | |
|---|---|
| Merge by compare-and-swap | Write back conditional on the row being unchanged and still unreserved. A losing merge affects no rows and retries |
| Claim by conditional update | `UPDATE ... SET reserved = true WHERE id = ? AND reserved = false`, checking affected rows |
| Push only when nothing is pending | A job already working the queue picks up what arrives while it runs |
| Job loops over every pending row | Re-queues if a row arrives as it exits |
| Release reservations older than the queue's `ttr` | Nothing frees a row when a worker is killed rather than caught, so that work was stranded silently |

We first tried dropping the merge and inserting a row per save. That made the write path much faster, 0.26s against 7.0s for 1200 saves, but it put `generateCatalogPrices()` on every row: 60 IDs in one call measured 0.025s against 0.081s as 60 single-ID calls. We kept the merge.

We also tried claiming every pending row up front. Killing a worker mid-drain left 200 rows reserved with nothing to free them, so we kept claiming one row at a time.

Our changes don't include any schema change, or `SKIP LOCKED` or `RETURNING`.

## Verification

Postgres, `runQueueAutomatically` off, no workers running. 
Six processes, 100 `createCatalogPricingJob()` calls each with a distinct purchasable ID.

| | 5.7.2 | Patched |
|---|---|---|
| Queue rows | 1 | 1 |
| IDs preserved | 600 | 600 |
| Errors | 0 | 0 |
| **Jobs pushed** | **600** | **3** |

**Mutex contention.** Holding `catalogpricingqueue` and then saving throws `RuntimeException: Unable to acquire the catalog pricing queue mutex.` on 5.7.2, succeeds patched.

**Lost updates.** Compare-and-swap against an unguarded read-modify-write of the same row, six processes merging 30 IDs each:

| | Expected | Survived |
|---|---|---|
| Read-modify-write | 180 | 66 |
| Compare-and-swap | 180 | **180** |

**Reservation.** Eight processes against 400 seeded rows: 400 claims, 400 distinct, 0 double-claimed, 0 left unreserved.

**Crash.** A worker that is killed after claiming leaves just its 1 row reserved; others drain the rest, and the expired-reservation release recovers it.

### DIFF
--- a/src/queue/jobs/CatalogPricing.php
+++ b/src/queue/jobs/CatalogPricing.php
@@ -7,8 +7,10 @@
 
 namespace craft\commerce\queue\jobs;
 
+use Craft;
 use craft\commerce\Plugin;
 use craft\commerce\records\CatalogPricingQueue as CatalogPricingQueueRecord;
+use craft\helpers\Queue as QueueHelper;
 use craft\queue\BaseJob;
 
 class CatalogPricing extends BaseJob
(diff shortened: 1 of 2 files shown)


sjcallender:
I should add that we really only see the error on staging/production. Locally, it finishes too fast, so our test forces the issue.

lukeholder:
Thanks for the PR, we will review and give feedback soon. Thanks!

sjcallender:
Thanks @lukeholder. While we have hopes for you all restructuring how all this works, this PR stays closer to the current setup. It is AI-assisted output for sure, but human-guided in shape and thought. Hope there's something valuable in here to help alleviate our builds' issues.

PR #4362, opened 2026-09-07
[5.x] Cache control-panel order status badge counts

boboldehampsink:
### Description

The control-panel order index already loads its status badges asynchronously, but each badge request reruns `getOrderCountByStatus()` against the orders and elements tables. With a large order history, repeated index refreshes repeat the same expensive grouped count even when no relevant order state has changed.

Cache the grouped counts per store through Craft's configured cache component, without depending on Redis or another specific backend. Status handles and zero-count statuses are still resolved outside the cache, preserving the existing response structure and fresh status labels.

Invalidate affected stores when orders become completed or incomplete, completed orders change status or store, or completed orders are deleted or restored. Cart saves and unrelated order edits leave the cache intact. Invalidation runs after the database transaction ends, and reads inside a transaction bypass the shared cache to avoid sharing uncommitted counts. Existing global and order-type cache clearing remains supported.

A one-hour expiry provides fallback reconciliation for direct database updates that bypass the order lifecycle; these updates can leave badge counts stale until expiry unless they explicitly invalidate the counts.

### Related issues

No linked issue.

### DIFF
--- a/src/elements/Order.php
+++ b/src/elements/Order.php
@@ -2282,6 +2282,8 @@ class Order extends Element implements HasStoreInterface
             }
 
             $oldStatusId = $orderRecord->orderStatusId;
+            $oldStoreId = $orderRecord->storeId;
+            $wasCompleted = (bool)$orderRecord->isCompleted;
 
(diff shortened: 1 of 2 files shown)


lukeholder:
Thanks for the PR, we will review and give feedback soon. Thanks!

PR #4366, opened 2026-09-15
[5.x] Fix custom fields with visibility conditions not saving on orders/carts

### LINKED ISSUE #4198: [5.x]: Can't save custom order field with visibility condition
### Description
Context:
I have a custom order entry field called `Pickup in store` which should only be required and visible when the shipping method is `Pickup in a store`.
The entry field returns a list of stores.
This field is required.

Issue: 
I can only save the order in the control panel (by clicking "update order") if the `Pickup in store` field does not have an order condition. With the order condition it refreshes the page and removes the chosen store even though I selected one. Without the order condition it allows me to save the order.

### Steps to reproduce

1. Create a custom order entry field. In my testing the issues does not occur with other field types.

Image

2. Set the order condition in the order field layout and make it required

Image

3. Try to create a manual order in the control panel

### Expected behavior
The order should be saved with the selected entry field value.

### Actual behavior
I am stuck with this error.
Disabling the order condition "fixes" the issue.

Image

### Craft CMS version

5.8.21

### Craft Commerce version

5.5.2

### PHP version

8.2.29

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

lukeholder:
## Summary
Custom fields' values were set before order/cart attributes like `shippingMethodHandle`, so field layout visibility conditions based on those attributes evaluated against stale state and the submitted value got dropped. Reordering so attributes are set first fixes it.

Clicking 'update order' is still required to refresh the custom field UI. We hope to improve this in the future.

Fixes #4198


nfourtythree (approved):
Yea looks good to me. Along with what you have stated about still needing an "update order" click. Hopefully we can look at an improved implementation of that in the future.

